PR #20, opened 2014-06-27
FileCookieStore is implemented

abelozerov:
Have implemented Store that saves data to file in UTF8 format. On construction data is loaded from file and stored in idx property and on read operations works the same as standard MemoryCookieStore. On write operations it dumps idx to file.


stash:
Several out in the ecosystem as Store implementations, plus this base repo is now gone

Issue #19, opened 2014-06-27
Incorrect cookie date handling (expires dates) - maybe?


Sironfoot:
Boots.com website returns cookies with expires values that look like this:

````
Thu, 26-Jun-14 18:59:59 GMT
Thu, 01-Dec-94 16:00:00 GMT
````

Notice the years are 2 digits only. But when they are parsed with tough-cookie like so:

````javascript
var tough = require('tough-cookie');

console.log(tough.parseDate('Thu, 26-Jun-14 18:59:59 GMT'));
console.log(tough.parseDate('Thu, 01-Dec-94 16:00:00 GMT'));
````

...the output I get is:

````
Mon Jun 26 142000 19:59:59 GMT+0100 (BST)
Invalid Date
````

Note the year 142000. For an example page on Boots.com that returns lots of cookies check out the login page at https://www.boots.com/webapp/wcs/stores/servlet/LogonForm?storeId=10052&langId=-1&catalogId=11051&URL=http%3A%2F%2Fwww.boots.com%2F

Now, an actual browser, such as Chrome, will parse these dates correctly. So who's in the right here, Chrome or tough-cookie? Are 2 digit years valid dates in the cookie spec? Is tough-cookie just being too strict?

I came across this problem when scraping the Boots.com website with 'request' and raised a bug there https://github.com/mikeal/request/issues/948 . But since request uses tough-cookie they suggested raising the issue here.

lalitkapoor:
@stash any progress on this?

inikulin:
Fixed in #30. Released with tough-cookie@0.13.0.

PR #21, opened 2014-07-15
issue #19

egor-kanakin:
Fixed 2-digit year workaround.
One plus-sign added converts a 2-digit year-string to an integer before the year-value is incremented.


stash:
This was fixed in another branch and #19 is now fixed.

PR #30, opened 2015-03-12
Adopt IETF's test suite, fix spec and browser compliance errors

inikulin:
Hello!

We've decided to adopt tough-cookie for the jsdom (tmpvar/jsdom#1027). Therefore I've adopted [IETF's test suite](https://github.com/abarth/http-state) and fixed issues that arose.


domenic:
+1, would love to be able to have jsdom depend on something with a cute turtle logo.

apsavin:
Where is somebody who can merge it?)

JSBizon:
looks this important repository isn't supported anymore, maybe make sense create a fork? I did one for myself: https://github.com/JSBizon/tough-cookie

inikulin:
This is a plan B =)

JSBizon:
right, but owner of this repo is silent for a long time

domenic:
Someone should email https://www.npmjs.com/~goinstant and https://www.npmjs.com/~jstash to see if they'd be willing to transfer the npm package ownership at least

stash:
Hey folks.  I'm so, so sorry to have taken so long to respond.  Long story short (I can explain if you want) is that I lost control of this repo for legal reasons, but now I can maintain it again!

I will look over the PR.  Salesforce is currently requiring that contributors sign a CLA. If the PR looks good, I will need to get one from @inikulin and then I can publish to NPM.

lib/cookie.js
```diff
@@ -55,21 +61,10 @@ var PATH_VALUE = NON_CTL_SEMICOLON;
...
- * [fail if]
- * *  the hour-value is greater than 23,
- * *  the minute-value is greater than 59, or
- * *  the second-value is greater than 59.
- */
-var TIME = /(0?[0-9]|1[0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9])/;
-var STRICT_TIME = /^(0?[0-9]|1[0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9])$/;
+var DAY_OF_MONTH = /^(\d{1,2})[^\d]*$/;
+var TIME = /^(\d{1,2})[^\d]*:(\d{1,2})[^\d]*:(\d{1,2})[^\d]*$/;
+var MONTH = /^(Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec)/i;
```
stash: Is the removal of the `$` anchor here intentional?

lib/cookie.js
```diff
@@ -457,6 +474,8 @@ function parse(str, strict) {
        * context of the parsing.
        */
       if (!av_value || av_value.substr(0,1) != "/") {
+        c.path = null;
+        c.path = null;
```
stash: Duplicated assignment.

lib/cookie.js
```diff
@@ -486,6 +505,8 @@ function parse(str, strict) {
 
   // ensure a default date for sorting:
   c.creation = new Date();
+  //NOTE: high-resolution time for the sorting in runtime
+  c.hrCreation = hrTimeNow();
```
stash: Can you please explain what the goal is here? I can see what it does, but I don't understand the motivation or why it's important? Since it's a small thing, perhaps you could split this off into another PR?

My concern is that if some code manually assigns to `.creation`, there's no setter that changes `.hrCreation` to match. This would result in inconsistent sorting in `cookieCompare`.

lib/cookie.js
```diff
@@ -820,7 +852,7 @@ var CAN_BE_SYNC = [];
 CAN_BE_SYNC.push('setCookie');
 CookieJar.prototype.setCookie = function(cookie, url, options, cb) {
   var err;
-  var context = (url instanceof Object) ? url : urlParse(url);
+  var context = getCookieContext(url);
```
stash: nice extraction. thanks :)

lib/cookie.js
```diff
@@ -869,15 +901,12 @@ CookieJar.prototype.setCookie = function(cookie, url, options, cb) {
     cookie.domain = host;
   }
 
-  // S5.3 step 7: "Otherwise, set the cookie's path to the default-path of the
-  // request-uri"
-  if (!cookie.path) {
+  //S5.2.4 If the attribute-value is empty or if the first character of the
+  //attribute-value is not %x2F ("/"):
+  //Let cookie-path be the default-path.
+  if (!cookie.path || cookie.path[0] !== '/') {
```
stash: This introduces another case where `.pathIsDefault` gets set. Is that correct in your opinion? Does the fall-through logic they say in S5.3 step 7 still need to happen or does the code as you have it here already cover it?

lib/memstore.js
```diff
@@ -59,18 +59,19 @@ MemoryCookieStore.prototype.findCookies = function(domain, path, cb) {
...
-        if (!pathIndex) {
-          return;
-        }
-        for (var key in pathIndex) {
-          results.push(pathIndex[key]);
-        }
-      });
-    };
+       //NOTE: we should use path-match algorithm from S5.1.4 here
+       //(see : https://github.com/ChromiumWebApps/chromium/blob/b3d3b4da8bb94c1b2e061600df106d590fda3620/net/cookies/canonical_cookie.cc#L299)
```
stash: Could you please explain here why we should use that path-match implementation?

Tactical: if this is a TODO, then maybe that goes as a comment on `pathMatch`? Or are you saying that `matchRFC` doesn't sync up with how Chromium does it?

package.json
```diff
@@ -24,7 +24,7 @@
   },
   "main": "./lib/cookie",
   "scripts": {
-    "test": "vows test.js"
+    "test": "vows test/*_test.js"
```
stash: YES thank you for splitting these up!

stash:
@inikulin I'm in the middle of looking at the changes and, overall, I'm impressed.  I'm confident we can work through the minor issues.  Most importantly: THANK YOU for all of the hard work! This is great!

test/cookie_sorting_test.js
```diff
@@ -0,0 +1,66 @@
...
+        cookies.push(Cookie.parse("c=2; Domain=example.com; Path=/pathA"));
+        cookies.push(Cookie.parse("d=3; Domain=www.example.com; Path=/pathA"));
+        cookies.push(Cookie.parse("e=4; Domain=example.com; Path=/pathA/pathB"));
+        cookies.push(Cookie.parse("f=5; Domain=www.example.com; Path=/pathA/pathB"));
+
+        // force a stable creation time consistent with the order above since
+        // some may have been created at now + 1ms.
+        var i = cookies.length;
+        cookies.forEach(function (cookie) {
+          cookie.creation = new Date(now - 100 * (i--));
```
stash: If `.hrCreation` gets left in, then this doesn't do anything since `.hrCreation` takes precedence over `.creation`. The end result is coincidentally the same, making this `forEach` loop redundant.

stash: Not saying it should be left in; we should still discuss it (and probably have some unit tests that cover it).

test/ietf_test.js
```diff
@@ -0,0 +1,96 @@
...
+          assert.strictEqual(actual, testCase.expected);
+        };
+      });
+    });
+
+  return {'Date': theVows};
+}
+
+vows
+  .describe('IETF http state tests')
```
stash: Again, seriously awesome that you're adding these test cases.  Thanks :)

lib/cookie.js
inikulin: You mean in month regexp? Yes, according to RFC:

>  month           = ( "jan" / "feb" / "mar" / "apr" /
                       "may" / "jun" / "jul" / "aug" /
                       "sep" / "oct" / "nov" / "dec" ) *OCTET

*OCTET - means that we can have any suffix. So, thing like `January` and even `JanWhatever` are acceptable. In Chromium source: https://github.com/ChromiumWebApps/chromium/blob/b3d3b4da8bb94c1b2e061600df106d590fda3620/net/cookies/cookie_util.cc#L125

lib/cookie.js
inikulin: Nice catch, thanks.

lib/cookie.js
inikulin: Previous implementation caused flickering of the IETF's ordering test. Cookies were added too fast and `Date()` resolution was not enough to provide the correct order. We can reset `hrCreation` if `creation` was set manually.

lib/cookie.js
inikulin: S5.2.4 cases are: `Path=;` and `Path=anything_except_slash`. S5.3.7 covers case when we don't have path attribute at all. In all those cases we should fallback to the default-path. So, yes this code will cover both S5.2.4 and S5.3.7.

lib/memstore.js
inikulin: This is according to the S5.4.1:
> The user agent MUST use an algorithm equivalent to the following
   algorithm to compute the "cookie-string" from a cookie store and a
   request-uri:
....
*  The request-uri's path path-matches the cookie's path.

BTW, seems like we can remove other matcher. I'll try it and will add another commit in case if  it works.

test/ietf_test.js
inikulin: You are welcome! =)

inikulin:
Weird, BSD date tests starts to fail after March 30th. Need to take a look at it.

inikulin:
Wow!
Setting system date to the March 31:
```js
$node
> var a = new Date(); a.setUTCMonth(3); a.getUTCMonth() === 3
false
```
Setting system date to the March 30:
```js
$node
> var a = new Date(); a.setUTCMonth(3); a.getUTCMonth() === 3
true
```
Am I missing something?

inikulin:
Ok, now we need to consider what to do with the `hrCreation`

lib/cookie.js
domenic: Naive outsider question but why not use an incrementing counter in module scope?

inikulin: @domenic it will work as well, but hrTime looks more expressive IMHO.

inikulin:
Ok, I think some tests required for the latest commit to be sure that we are not breaking `cookieCompare` if `cookie.creation` was modified by the user. But, I need to go now, so I'll add them tomorrow.

stash:
There's another issue open for the date thing.  I have a fix waiting in the wings that I'll merge in after this one.

lib/cookie.js
stash: :+1: great.

lib/memstore.js
stash: OK. A separate PR sounds good just to keep this big one reasonable.

stash:
re-reviewed; looking great still.

domenic:
Anything left before this can be merged?

stash:
I will cut a release for this tomorrow. I'm still sorting out NPM ownership so please bear with me.

stash:
Still working on this, stay tuned.

stash:
Published as v0.13.0

PR #37, opened 2015-04-27
Get rid of strict mode (closes #34, fixes #9)

### LINKED ISSUE #9: Cookie.parse should be directly usable from map
`Cookie.parse` was changed a while ago to accept a second `strict` boolean parameter.  This makes it no longer possible to pass it directly to `Array#map`, since that passes a integer index as the second parameter (see #7).

As a breaking API change, make `Cookie.parse` always use non-strict mode and `Cookie.parseStrict` always use strict mode.

### LINKED ISSUE #34: Remove Strict Mode
Via @inikulin: 

> The thing that bothers me most is the strict mode. If enabled it doesn't pass IETF's tests, therefore currently non-strict mode is stricter than "strict" mode. So, I suggest to get rid of "strict" mode. I can issue a PR if you are OK with it.

In the interest of simplicity and compliance, we should remove this mode prior to 1.0.  This means an API breaking change, so let's maintain it on a "1.0-pre" branch in case we need to apply fixes to 0.x (currently 0.13)

inikulin:


stash-sfdc:
Looks good to me

stash-sfdc:
Oops, missed a few mentions of strict mode. Will patch

PR #42, opened 2015-05-25
Serializable

stash-sfdc:
WIP until documentation is done. The branch name is misleading; I intend for this to be for all store types, not just `MemoryCookieStore` (which is what I started out with).

---

Add a `getAllCookies` interface to `CookieStore` so that stores can export all cookies.  In the case of stores where this is intractable or undesirable, they can elect to not implement `getAllCookies`.

The serialized format is vaguely:

```js
  var serialized = {
    // The version of tough-cookie that serialized this jar. Generally a good
    // practice since future versions can make data import decisions based on
    // known past behavior. When/if this matters, use `semver`.
    version: 'tough-cookie@'+VERSION,

    // add the store type, to make humans happy:
    storeType: type,

    // CookieJar configuration:
    rejectPublicSuffixes: !!this.rejectPublicSuffixes,

    // this gets filled from getAllCookies after calling `toJSON` on them:
    cookies: []
  };
```

Importing the cookies repeatedly calls `store.putCookie()`, so this should work for all `CookieStore` implementations, in theory.  `Cookie.fromJSON` is called on each item in the `cookies` Array before being passed to `.putCookie()`.


stash-sfdc:
@lalitkapoor @inikulin @apsavin initial thoughts? Did I miss some way this might break existing clients? Can someone spot-check this with `request` and/or `domjs`? 

I'd like to polish this up with documentation, etc. if yall are good with it.

test/cookie_to_json_test.js
```diff
@@ -123,8 +123,8 @@ vows
       },
       "null": {
         topic: function(f) { return f(null) },
-        "looks good": function(str) {
-          assert.match(str, /"maxAge":null/);
+        "absent": function(str) {
+          assert.match(str, /(?!"maxAge":null)/); // NB: negative RegExp
```
stash-sfdc: Removed this since it === what's in the prototype.

lib/cookie.js
```diff
@@ -473,23 +474,36 @@ function fromJSON(str) {
...
+    {
+      continue; // leave as prototype default
     }
+
     if (prop === 'expires' ||
         prop === 'creation' ||
         prop === 'lastAccessed')
     {
-      c[prop] = obj[prop] == "Infinity" ? "Infinity" : new Date(obj[prop]);
+      if (obj[prop] === null) {
```
stash-sfdc: Added this check because `new Date` was failing, obviously. ;)

lib/cookie.js
```diff
@@ -603,13 +619,13 @@ Cookie.prototype._initialCreationTime = null; // Used to determine if cookie.cre
...
 Cookie.prototype.lastAccessed = null; // Date when set
 
-var cookieProperties = Object.freeze(Object.keys(Cookie.prototype).map(function(p) {
-  if (p instanceof Function) {
-    return;
-  }
-  return p;
-}));
-var numCookieProperties = cookieProperties.length;
+Cookie.serializableProperties = Object.keys(Cookie.prototype)
```
stash-sfdc: I think it's OK to have this non-frozen, documented, and therefore customizable. What do yall think?

lib/cookie.js
```diff
@@ -620,6 +636,43 @@ Cookie.prototype.inspect = function inspect() {
     '"';
 };
 
+Cookie.prototype.toJSON = function() {
+  var obj = {};
+
+  var props = this.serializableProperties || Cookie.serializableProperties;
```
stash-sfdc: Maybe folks want to store additional things on the `Cookie` itself, and this would allow them to add fields.

TODO: Really this is probably a set-union of the custom + standard properties, or just allow users to amend `Cookie.serializableProperties`.

lib/memstore.js
```diff
@@ -116,21 +118,21 @@ MemoryCookieStore.prototype.putCookie = function(cookie, cb) {
   cb(null);
 };
 
-MemoryCookieStore.prototype.updateCookie = function updateCookie(oldCookie, newCookie, cb) {
+MemoryCookieStore.prototype.updateCookie = function(oldCookie, newCookie, cb) {
```
stash-sfdc: If you're curious why I used to do this all the time... let's just say node 0.4.x sucked ;)

stash-sfdc: ... and/or i was naive. ;)

lib/cookie.js
lalitkapoor: Hmm, will deserialization add back any of the additional fields? It looks like deserialization is using `fromJSON` and that only sets values for keys is `Cookie.serializableProperties` - I think we may lose additional fields on deserialization if I'm not mistaken.

lalitkapoor:
@stash-sfdc looks great! - left one comment above

lalitkapoor:
will give it a whirl with request

lib/cookie.js
stash-sfdc: You're not mistaken. I think I'll keep it simple and get rid of `this.serializableProperties`.

lalitkapoor: :+1:

lalitkapoor:
all request tests pass. Clone the memory store jar is frequent request in request - this is awesome. :+1:

inikulin:
LGTM :+1:

stash-sfdc:
@apsavin good point. Old habits die hard so I really appreciate you pointing these out!

README.md
```diff
@@ -149,12 +150,14 @@ Properties
...
   * _extensions_ - `Array` - any unrecognized cookie attributes as strings (even if equal-signs inside)
+  * _creation_ - `Date` - when this cookie was constructed
+  * _creationIndex_ - number - set at construction, used to provide greater sort precision (please see `cookieCompare(a,b)` for a full explanation)
 
 After a cookie has been passed through `CookieJar.setCookie()` it will have the following additional attributes:
 
   * _hostOnly_ - boolean - is this a host-only cookie (i.e. no Domain field was set, but was instead implied)
   * _pathIsDefault_ - boolean - if true, there was no Path field on the cookie and `defaultPath()` was used to derive one.
-  * _created_ - `Date` - when this cookie was added to the jar
+  * _creation_ - `Date` - **modified** from construction to when the cookie was added to the jar
```
stash-sfdc: Ack, this was always `creation`, never `created` :(

stash-sfdc:
OK, revisions to retain creation sort order (see a5dfae6) and documentation are in.

Is anyone going to get mad that custom properties are being dropped in `.toJSON()`? If so, I guess semver says this should go to 2.0.0 since it's an API-breaking change.

lalitkapoor:
I think we should follow semver, and since it's breaking let's do the major bump, followed by a release party :P

apsavin:
Why somebody will want to use custom properties?

inikulin:
Ok, I think we can merge now (but we need to bump version to 2.0.0 first). @stash Something else is left to do?

stash-sfdc:
Nothing left, I think (thanks for the nudge). I'll re-review and tag 2.0 if it's good.

stash-sfdc:
Alright, party like it's `Fri, 31 Dec 1999 23:59:59 GMT` ;)

https://twitter.com/jstash/status/608759537936367617

inikulin:
:tada:

lalitkapoor:
Image of Cookie Monster Party

stash-sfdc:
haha, brilliant! thanks @lalitkapoor :)

Issue #10, opened 2014-01-06
Parse "bare" cookies in non-strict mode


stash:
Some cookies with an empty value can look like `name` instead of `name=`.  These should be parse-able in non-strict mode and round-tip when serializing to strings.

stash:
Triage: should investigate if this is still an issue, sync up with the whatwg reference linked above?

stash:
Related to this: check out http://inikulin.github.io/cookie-compat/ !

ruoho:
Correctly parsing cookies with empty / no key was merged in https://github.com/salesforce/tough-cookie/pull/49. Let's create a new issue if we want to investigate the whatwg reference.

PR #49, opened 2015-08-01
Enable parsing cookies with an empty or no key

Sebmaster:
This is basically a proposal for fixing #10 since I came across an instance where I'd need this.

I'm not sure about the handling of the "re-introduced" strict parameter. In order to not change behaviour when not given I set it to default to true for now, but I guess we'd need a better solution here?

### DIFF
--- a/lib/cookie.js
+++ b/lib/cookie.js
@@ -49,7 +49,7 @@ var DATE_DELIM = /[\x09\x20-\x2F\x3B-\x40\x5B-\x60\x7B-\x7E]/;
 // From RFC6265 S4.1.1
 // note that it excludes \x3B ";"
 var COOKIE_OCTET  = /[\x21\x23-\x2B\x2D-\x3A\x3C-\x5B\x5D-\x7E]/;
-var COOKIE_OCTETS = new RegExp('^'+COOKIE_OCTET.source+'$');
+var COOKIE_OCTETS = new RegExp('^'+COOKIE_OCTET.source+'+$');
 
 // Double quotes are part of the value (see: S4.1.1).
 // '\r', '\n' and '\0' should be treated as a terminator in the "relaxed" mode
--- a/lib/cookie.js
+++ b/lib/cookie.js
@@ -51,12 +51,16 @@ var DATE_DELIM = /[\x09\x20-\x2F\x3B-\x40\x5B-\x60\x7B-\x7E]/;
 var COOKIE_OCTET  = /[\x21\x23-\x2B\x2D-\x3A\x3C-\x5B\x5D-\x7E]/;
 var COOKIE_OCTETS = new RegExp('^'+COOKIE_OCTET.source+'+$');
 
+var CONTROL_CHARS = /[\x00-\x1F]/;
+
 // Double quotes are part of the value (see: S4.1.1).
 // '\r', '\n' and '\0' should be treated as a terminator in the "relaxed" mode
 // (see: https://github.com/ChromiumWebApps/chromium/blob/b3d3b4da8bb94c1b2e061600df106d590fda3620/net/cookies/parsed_cookie.cc#L60)
 // '=' and ';' are attribute/values separators
 // (see: https://github.com/ChromiumWebApps/chromium/blob/b3d3b4da8bb94c1b2e061600df106d590fda3620/net/cookies/parsed_cookie.cc#L64)
-var COOKIE_PAIR = /^([^=;]+)\s*=\s*(("?)[^\n\r\0]*\3)/;
+var COOKIE_PAIR = /^(([^=;]+))\s*=\s*(("?)[^\n\r\0]*\3)/
+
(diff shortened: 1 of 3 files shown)


stash-sfdc:
Hi @Sebmaster. Thanks for the pull request! If you're interested, it needs a bit of work. Please note that SFDC requires a CLA to be signed in order to accept any PR (either by you or your employer).

We got rid of a similar `strict` flag when we moved from 1.x to 2.x, so I'd be hesitant to reintroduce it on the parse method. The problem is that folks do stuff like `arrayOfStrings.map(Cookie.parse)` which leads to some odd behavior since the function passed to [`Array#map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) has the signature `function(currentValue, index, array)`. Since 0 is implicitly `false`, the very first string in the array gets parsed in "non-strict" and the subsequent, non-0 indexes are "strict". So, I'd like to please see a different API implementation. Perhaps tough-cookie needs some kind of global configuration options? Maybe change from "Class-static" to a "default-Factory-method" pattern (similar to how `request` and node's `http.Agent` allocate a default global to use)?

I'd like to try to maintain RFC compliance if possible, so unless this is how the RFC says to parse it (I honestly haven't looked yet), I don't think this feature should be "on" by default. You've implemented it as "off" by default, which is great, but I wanted to be clear that we can't just switch to `LOOSE_COOKIE_PAIR` all the time. Specifically, can you confirm that in the case of both cookie-strings `abc` and `=abc` that it's got no key but value `abc`?

Thanks again for taking the time to contribute this PR :100:

Sebmaster:
> So, I'd like to please see a different API implementation.

I actually thought of another method in the meanwhile: Make the parameter an options object instead and check for explicit existence of a strict property in that object. If the given param is either not an object, or that key doesn't exist, we just ignore it.

> Specifically, can you confirm that in the case of both cookie-strings abc and =abc that it's got no key but value abc?

So, according to the RFC this is not actually a valid cookie - I think. However, all browsers can properly parse this cookie. Sending a `Set-Cookie` equal to "=abc", leads browsers to send a `Cookie: abc` header back, so we can say `"abc" == "=abc"`. Since setting a cookie with `abc=` sets a different cookie (`abc=`), we can conclude that the name of the cookie is empty and the value is abc in both parsing cases.

> Please note that SFDC requires a CLA to be signed in order to accept any PR (either by you or your employer).

Where can I find this CLA?

inikulin:
@stash-sfdc @Sebmaster the thing is that nobody fully implements RFC spec :tada:  As far as I remember even IETF tests violate it at some cases in favor of common browser behavior. So, if it doesn't break existing tests and if this behavior is consistent across browsers, I'd rather keep it as non-optional.

stash-sfdc:
@inikulin OK, agreed: if it's both IETF-test-compliant and consistent, then it should be the default behavior.

Sebmaster:
So we remove the non-strict param altogether? As far as I remember this starts breaking the (IETF part of the) test suite, so what should I do about that?

stash-sfdc:
@Sebmaster please see https://github.com/forcedotcom/idecore/wiki/Contributing-Code regarding the CLA.

stash-sfdc:
@Sebmaster OK, i follow the key-versus-value logic. Thanks for explaining that clearly!

If on-by-default breaks the suite, then perhaps an options Object as the second parameter (that handles the Array#map use-case) would be best.

inikulin:
@Sebmaster I have a feeling what we can make both, we just need a more sophisticated regexp. No facts, just an assumption :walking:

Sebmaster:
If I remember correctly the IETF suite explicitly tests to disallow the format we want to allow here, so I doubt it. Will look into it though.

Sebmaster:
Moved to options object and also added another test I found in the IETF tests which looked like a pretty interesting scenario to support (verified in browsers).

Sebmaster:
CLA is done too now.

stash-sfdc:
Confirm CLA accepted! Welcome @Sebmaster!

lib/cookie.js
```diff
@@ -885,7 +899,7 @@ CookieJar.prototype.setCookie = function(cookie, url, options, cb) {
 
   // S5.3 step 1
   if (!(cookie instanceof Cookie)) {
-    cookie = Cookie.parse(cookie);
+    cookie = Cookie.parse(cookie, !options || !options.loose);
```
stash-sfdc: Should this pass `{loose: options.loose}` or maybe just `options` entirely? On a hunch this might require code coverage -- try adding `jar.setCookieSync('=a', url)` and `jar.setCookieSync('=a',url,{loose:true})` and make sure those behave as expected?

Sebmaster: Nice catch, fixed and added tests for it.

stash-sfdc:
@Sebmaster looks good to me. Ready to be merged?

Sebmaster:
Yep, should be good. :tada:

PR #56, opened 2015-10-03
Implement loose mode for cookie jars

Sebmaster:
Forgot that this might be needed too, especially if tough-cookie is used in conjunction with `request`, which doesn't set loose by default.

I was thinking about doing an options object as a param instead, but I don't think that the already existing option can be replaced in a backwards-compatible way, so we could only add an options object in addition. Not sure about what to do here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,15 +237,15 @@ if (cookie.validate() === true) {
 
 Exported via `tough.CookieJar`.
 
-### `CookieJar([store],[rejectPublicSuffixes])`
+### `CookieJar([store],[options])`
 
 Simply use `new CookieJar()`.  If you'd like to use a custom store, pass that to the constructor otherwise a `MemoryCookieStore` will be created and used.
 
-### Properties
+The `options` object can be omitted and can have the following properties:
 
-CookieJar object properties:
-
-  * _rejectPublicSuffixes_ - boolean - reject cookies with domains like "com" and "co.uk" (default: `true`)
+  * _rejectPublicSuffixes_ - boolean - default `true` - reject cookies with domains like "com" and "co.uk"
+  * _looseMode_ - boolean - default `false` - accept malformed cookies like `bar` and `=bar`, which have an implied empty name.
+    This is not in the standard, but is used sometimes on the web and is accepted by (most) browsers.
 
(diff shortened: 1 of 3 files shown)


Sebmaster:
This might not be needed if https://github.com/request/request/pull/1811 gets merged, but just in case we want to provide this functionality anyways...

stash-sfdc:
Nice, yes, this is needed.

If we're going to change the signature of `CookieJar`'s constructor, I think it should be `(store, options)` instead of `(store, rejectPublicSuffixes, looseMode)`. It should have been that way originally, in hindsight. Either way it's a SemVer MINOR bump. Can be backwards compatible, e.g.:

```js
function CookieJar(store, options) {
  if (typeof options === 'boolean') {
    options = { rejectPublicSuffixes: options };
  }
// ...
  this.rejectPublicSuffixes = options.rejectPublicSuffixes
```

Could you also document the change in `README.md`, please and thank you?

Sebmaster:
Implemented. It's not quiiite backwards compatible since before you didn't have to pass a boolean arg and it'd still work.

Locally it seems like master failed before this, which also now causes the PR to fail. This seems to not be a case on Travis though?!
There's something weird about this test format though, so I'm not quite sure why that is. Seems like there's an Error object in the second argument instead of the first one [here](https://github.com/SalesforceEng/tough-cookie/blob/master/test/api_test.js#L392).

stash-sfdc:
Did you update vows locally? I bumped to using 0.8.1 on master.

Aside: **TODO**: stop using vows :wink:. It was a good choice at the time tough-cookie was created, but I find the simplicity/stability of mocha (or even TAP) to be qualitatively, and opinionatedly, better.

Sebmaster:
Yep, that was it. Successful here too then :+1:

stash-sfdc:
Oops, hold on, didn't push something so will re-merge.

stash-sfdc:
ok, should be good. will publish as 2.2.0 shortly

stash-sfdc:
Published! Thanks again @Sebmaster, you rock!

Sebmaster:
Nice, hopefully I can finally put this 2 year old bug I found to bed :grinning:

PR #67, opened 2016-05-06
S8.5 Ignore ports in cookie domains

alubbe:
I was using the request module for my test suite to hit a few different services running on the same host, but under different ports. According to the spec (8.5), ports should be entirely ignored, but tough-cookie includes them.
This PR attempts to fix that and adds a test case.
Unfortunately, it also breaks the DOMAIN0027 test, and I don't really understand why. Any feedback is most welcome!

### DIFF
--- a/lib/cookie.js
+++ b/lib/cookie.js
@@ -432,7 +432,8 @@ function parse(str, options) {
       if (av_value) {
         // S5.2.3 "Let cookie-domain be the attribute-value without the leading %x2E
         // (".") character."
-        var domain = av_value.trim().replace(/^\./, '');
+        // S8.5 Ports should be ignored
+        var domain = av_value.trim().replace(/^\./, '').replace(/:\d{1,5}/, '');
         if (domain) {
           // "Convert the cookie-domain to lower case."
           c.domain = domain.toLowerCase();
--- a/test/ietf_data/parser.json
+++ b/test/ietf_data/parser.json
@@ -1005,7 +1005,9 @@
       "foo=bar; domain=home.example.org:8888"
     ],
     "sent-to": "http://home.example.org:8888/cookie-parser-result?domain0027",
-    "sent": []
+    "sent": [
+      { "name": "foo", "value": "bar" }
+    ]
   },
   {
     "test": "DOMAIN0028",
--- a/test/parsing_test.js
+++ b/test/parsing_test.js
@@ -130,6 +130,11 @@ vows
       assert.ok(c);
       assert.equal(c.domain, 'example.com');
     },
+    "port in domain": function() {
+      var c = Cookie.parse("a=b; domain=example.com:80");
+      assert.ok(c);
+      assert.equal(c.domain, 'example.com');
+    },
     "trailing dot in domain": {
       topic: function() {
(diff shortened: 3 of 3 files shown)


alubbe:
So DOMAIN0027 failed because, as I understand it, its expected behaviour was incorrect.
It set `foo=bar; domain=home.example.org:8888` on `http://home.example.org:8888/cookie-parser-result?domain0027`, which means that, ignoring ports everywhere, the cookie domain is actually just `home.example.org` and it gets used for every port, incl. `8888`. So instead of expecting no applicable cookie, it is my understanding that the test should expect `foo=bar` to be set successfully.

I've just committed an update to this PR and all tests are passing.

alubbe:
@stash-sfdc Any feedback?

inikulin:
S8.5 is non-normative. Can you point to to the exact parsing rule, which describes this change? Also, how browsers behave?

inikulin:
You got S8.5 wrong. Specified cookie domain should match request-host (defined in S2.3) as per S5.3.5. So, DOMAIN0027 works correctly. Meanwhile, cookie with domain `home.example.org` will be received by any `home.example.org:{port}` request origin, since we will use request-host for matching - this is what S8.5 desribes. Also, browser consistently implement this behavior - http://inikulin.github.io/cookie-compat/

alubbe:
I just checked both Firefox and Chrome. If you set a cookie on "localhost:3000", all future requests to just "localhost" get the same cookies.

inikulin:
`localhost` is a pub suffix. You need to test with different hostname

inikulin:
Also, it doesn't matter **where** you set it, it matters which `domain` property you use.

alubbe:
Upon further testing, it seems that `domain` properties with ports are ignored entirely, meaning the cookie is not set for any domain whatsoever. Whereas if you are on home.example.org:8888 and send a cookie without the `domain` property, it defaults to just "home.example.org", meaning it gets set for the current domain as well as all other ports on "home.example.org". Is this a correct summary of the behaviour / spec?

inikulin:
> Is this a correct summary of the behaviour / spec?

Yes. Also, you can specify `Domain` property, but it shouldn't contain port. Then any URL with any port but with hostname matching `Domain` will receive cookie.

alubbe:
Thank you. That last part was clear to me.
> but it shouldn't contain port

I was not aware that browsers declare a cookie invalid if the domain contains a port. I've just checked tough-cookie's implementation. The cookie does not get discarded but saved as "domain:port". However, this cannot be reached because `getCookieContext()` discards the port property, so the cookie is unreachable. Same outcome, but slightly different handling.

inikulin:
> I was not aware that browsers declare a cookie invalid if the domain contains a port.

They don't declare it invalid actually and they get into storage (as per spec). Such cookie just can't be fetched by any URL. I'm not sure if browsers optimize this case and don't store such cookies (haven't looked into their cookie related code for a quite long time). But it's all just implementation details. From user land perspective you can tell that they are "ignored"

alubbe:
Yes, that's clear to me now. Thanks again!
Just fyi, both chrome and firefox optimize and do not store them.

PR #71, opened 2016-08-24
Dockerized build/test environment for tough-cookie. 

awaterma:
Added in Istanbul code coverage reporting. A few changes to the .gitignore and the package.json as well (for the dependency and running coverage). I think this has everything we discussed. :) Please do comment and let me know if there is anything either of you would like changed/modified/updated!


apsavin:
Sorry, but why all the things in one commit?

apsavin:
Why do we need such a docker file into a repo? I mean, why it goes with vim? Very strange decision IMO. You can edit files locally with any editor and use docker only for tests. And why it's not in `.npmignore`?

package.json
```diff
@@ -58,13 +58,15 @@
   ],
   "scripts": {
     "suffixup": "curl -o public_suffix_list.dat https://publicsuffix.org/list/public_suffix_list.dat && ./generate-pubsuffix.js",
-    "test": "vows test/*_test.js"
+    "test": "vows test/*_test.js",
+    "coverage": "node node_modules/istanbul/lib/cli.js cover vows test/*_test.js"
```
awaterma: I think that requires a global install of istanbul, which is what I was trying to avoid. Is there another way to get an alias in the local project just through NPM and Node? I'm open to your suggestion; I was just trying to avoid errors on initial run of "npm run coverage".

awaterma:
I was being a bit too hesitant with .gitignore; thank you, I'll fix that.

I tried to split out the docker compositions so I could work in the container and use local settings; but yes, you are right, one can install and test in the container and edit on the outside. I'll make another change there and keep my local setup for my own work. In terms of why are the two in one commit, it's a simple changelist, and since its my first pull request, I thought I'd just gather up what I was doing around build and test into one pull request.

Appreciate your comments. :)

package.json
apsavin: npm adds `./node_modules/.bin` directory to PATH. I think, after local installation of istanbul you'll find `./node_modules/.bin/istanbul`, so you'll be able to use just `istanbul` in your npm scripts.

awaterma:
I have made quite a few changes based on your comments; users that don't use the docker image for build/test will have to globally install istanbul or use npm link if they want coverage. Thanks for the feedback; let me know what you think!

awaterma:
@apsavin -- please let me know if I should add a .npmignore for the Docker changes. NPM says that it relies on .gitignore when there is no .npmignore present; but if you would like me to be explicit, I'm willing to make another change. NPM on .npmignore/.gitignore:

https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package

awaterma:
I've tweaked on the simple request quite a bit; please let me know if you'd like me to close this and restructure my commits into one commit so the repo stays clean. I would then open a better/cleaner pull request. Thanks for the comments and questions.

awaterma:
I'm going to close this pull request and break it out into:

a) Istanbul coverage 
b) Dockerized setup (with .npmignore)

PR #73, opened 2016-09-27
Fixed Travis path after repository migration.

awaterma:
I think this is all of the modifications that need to be made to the repo to work with Travis after the migration. There may be more to this though; so please pass on some feedback.


awaterma:
@anish -- I think this is all that was required after the migration. Please let me know if you think I"m missing anything here.

anish:
Nope, that was all. Thanks for the follow up ! (however your team members still don't seem to have accepted their invites)

PR #91, opened 2017-08-12
Push expiry date further into the future

tomhughes:
This test had started failed because we have now passed the specified expiry date, so push it further into the future.


stash:
Looks like this or something similar has been incorporated already.

Issue #92, opened 2017-09-05
Vulnerable Regular Expression


cristianstaicu:
The following regular expression used for parsing the cookie is vulnerable to ReDoS:
```js
/^(([^=;]+))\s*=\s*([^\n\r\0]*)/
```
The slowdown is moderately low: for 50.000 characters around 2.5 seconds matching time. However, I would still suggest one of the following: 
- remove the regex,
- limit the number of characters that can be matched by the repetition,
- limit the input size.
I noticed there is another bug report regarding the correctness of this regular expression.

If needed, I can provide an actual example showing the slowdown.

adamwdennis:
Nice find. An actual example will be helpful in verification of a fix.

cristianstaicu:
Sure, here it is: 
```js
function genstr(len, chr) {
    var result = "";
    for (i=0; i<=len; i++) {
        result = result + chr;
    }
    return result;
}

var start = process.hrtime();
var tough = require('tough-cookie');
var str =  "x" + genstr(50000, ' ') + "x"; 
var Cookie = tough.Cookie;
var cookie = Cookie.parse(str);
var end = process.hrtime(start);

console.info("Execution time (hr): %ds %dms", end[0], end[1] / 1000000);
```

evilpacket:
@cristianstaicu remember to start your timer right before your parse operation (i.e. after your require and string generation) for accurate time measurements. Given that node limits header size to 80kb the dos is limited to about 7.3 seconds. At Node Security we consider anything over 1 second to be a valid issue.

eddieajau:
Just fyi, [nsp](https://github.com/nodesecurity/nsp) has this on their radar so anyone with `nsp` in their CI build pipeline will now be experiencing failing builds.

gemal:
https://nodesecurity.io/advisories/525

tkalfigo:
Indeed. Got the nsp error from our pre-commit hook. But is there a way to overcome it temporarily until a fix for tough-cookie is in place? I'd like to avoid of course disabling entirely the nsp check.

Hellenic:
You could [add an exception](https://github.com/nodesecurity/nsp#exceptions) to your `.npmrc` for that

subesokun:
Please fix as our builds are failing and I don't want to whitelist the advisory

gemal:
add the following to a `.nsprc` file:
```
{
  "exceptions": ["https://nodesecurity.io/advisories/525"]
}
```

[...]

PR #94, opened 2017-09-21
fix: ReDoS in cookie pair regexp

grnd:
Limiting whitespaces to 512. While the RFC does not specify a limit on number of whitespace in key, it's unreasonable to have thousands of them.

### DIFF
--- a/lib/cookie.js
+++ b/lib/cookie.js
@@ -58,7 +58,7 @@ var CONTROL_CHARS = /[\x00-\x1F]/;
 // (see: https://github.com/ChromiumWebApps/chromium/blob/b3d3b4da8bb94c1b2e061600df106d590fda3620/net/cookies/parsed_cookie.cc#L60)
 // '=' and ';' are attribute/values separators
 // (see: https://github.com/ChromiumWebApps/chromium/blob/b3d3b4da8bb94c1b2e061600df106d590fda3620/net/cookies/parsed_cookie.cc#L64)
-var COOKIE_PAIR = /^(([^=;]+))\s*=\s*([^\n\r\0]*)/;
+var COOKIE_PAIR = /^(([^=;]+))\s{0,512}=\s*([^\n\r\0]*)/;
 
 // Used to parse non-RFC-compliant cookies like '=abc' when given the `loose`
 // option in Cookie.parse:


grnd:
That test was failing before the change. will look into that, and add one for the ReDoS as well

inikulin:
@grnd test fail due to https://github.com/abarth/http-state/issues/18. We just need to update it.

inikulin:
@grnd Have you signed CLA?

lib/cookie.js
```diff
@@ -58,7 +58,7 @@ var CONTROL_CHARS = /[\x00-\x1F]/;
 // (see: https://github.com/ChromiumWebApps/chromium/blob/b3d3b4da8bb94c1b2e061600df106d590fda3620/net/cookies/parsed_cookie.cc#L60)
 // '=' and ';' are attribute/values separators
 // (see: https://github.com/ChromiumWebApps/chromium/blob/b3d3b4da8bb94c1b2e061600df106d590fda3620/net/cookies/parsed_cookie.cc#L64)
-var COOKIE_PAIR = /^(([^=;]+))\s*=\s*([^\n\r\0]*)/;
+var COOKIE_PAIR = /^(([^=;]+))\s{0,512}=\s*([^\n\r\0]*)/;
```
inikulin: It would be nice to leave a comment for future generations about why we need this limitation.

inikulin: And we need a regression test for this.

stewartjarod:
fixes #92

lib/cookie.js
arlolra: It looks like `trim()` is called on the captured key/val anyways, so why bother with the `\s*` altogether?  It's already in the capture group `[^=;]`.

arlolra: Also, note that `LOOSE_COOKIE_PAIR` seems to have the same problematic pattern.

stash-sfdc:
For the other ReDoS, we were able to write a test that demonstrated the bug. Would be great to see that as well here. Would be good to also do a systematic review for other ReDoS sites :(

I'll spend some time on this today. @grnd if you don't want to sign the CLA I can reject this PR and develop an alternative fix. Up to you.

alexanderbird:
I can't sign the CLA, but here are regression tests (adapted from the example given in issue 92 to reproduce the vulnerability) for the COOKIE_PAIR and LOOSE_COOKIE_PAIR. They fail without @grnd's fix, and pass if you apply the fix to both regexes. 

@grnd, you may have already written the regression tests, but in case you didn't, feel free to use this. 

`  .addBatch({
    "ReDoS vulnerabilities": {
      topic: function() {
        function genstr(len, chr) {
          var result = "";
          for (var i=0; i<=len; i++) {
            result = result + chr;
          }
          return result;
        }

        var NS_PER_SEC = 1e9;
        var maliciousString = "x"  + genstr(50000, ' ') + "x";

        return {
          maliciousString,
          NS_PER_SEC
        };
        this.callback();
      },
      "cookie pair regex is not vulnerable to ReDoS attacks": function(t) {
        var start = process.hrtime();
        var cookie = Cookie.parse(t.maliciousString);
        var end = process.hrtime(start);

        var elapsedSeconds = end[0] + (end[1] / t.NS_PER_SEC);
        assert.ok(elapsedSeconds < 0.25, "elapsedSeconds must be less than 0.25");
      },
      "loose cookie pair regex is not vulnerable to ReDoS attacks": function(t) {
        var start = process.hrtime();
        var cookie = Cookie.parse(t.maliciousString, { loose: true });
        var end = process.hrtime(start);

        var elapsedSeconds = end[0] + (end[1] / t.NS_PER_SEC);
        assert.ok(elapsedSeconds < 0.25, "elapsedSeconds must be less than 0.25");
      }
    }
  })`

westy92:
Prefer [chr.repeat(len)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat) over the custom `genstr(len, chr)`.

stash-sfdc:
Going to close this in favor of #97. Thank you @alexanderbird @grnd @westy92

PR #100, opened 2017-09-26
No RexExp parser

stash-sfdc:
I have nothing against RegExps in general, but for security conscious code, there's too many pitfalls. 

This pull-request is about converting from using mostly RegExp to using mostly native JavaScript code. The main cookie parser and the RFC-compliant date parser are most affected.

When tough-cookie was originally written, the RegExps were used because it seemed easier to maintain. This pull request certainly increases the code complexity, but does so in the hopes that maintaining the RegExps would have been more time-consuming in the long run.

As a bonus, it seems that the pure JavaScript version is about 1 to 2% faster than the RegExp one, based on the time it takes to run the entire tough-cookie test suite (7.0s versus 6.9s now).

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,7 +134,7 @@ else
   cookies = [Cookie.parse(res.headers['set-cookie'])];
 ```
 
-_Potentially non-standard behavior:_ currently, tough-cookie will limit the number of spaces before the `=` to 256 characters.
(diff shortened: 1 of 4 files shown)


stash:
Integration tested `request`, `jsdom`, and `zombie`.

awaterma:
Looks good. Thanks for merging.

PR #106, opened 2018-02-25
Replace public suffix implementation with `psl`.

stash:
Remove both the query function and list generator. Also remove the copy of the public suffix list itself.

See also https://github.com/wrangr/psl/issues/5 -- it makes sense to collaborate on auto-updating too!


stash:
Rebasing onto current master

PR #107, opened 2018-04-11
Added NYC coverge and reporting to tough-cookie.

awaterma:
Pulled in NYC (Istanbul 2.0) into our package.json. Did a few runs with the local package, set it up to do html based coverage reporting (through vows).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ node_modules/
 .*.sw[nmop]
 npm-debug.log
 docker-compose.override.yml
+.nyc_output
+coverage
--- a/.npmignore
+++ b/.npmignore
@@ -1,2 +1,4 @@
 docker-compose.yml
 Dockerfile
+.nyc_output
+coverage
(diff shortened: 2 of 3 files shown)


stash:
The overall coverage also looks pretty good. 

image

stash:
Looking through `cookie.js` there are both some trivial and important-to-test branches that aren't getting tested for. I'd like to shoot for 90+% if we can. However, we should at the very least not regress in coverage now ;)

awaterma:
Definitely agreed on keeping coverage up and working to extend the numbers
where possible. :)

There's a configuration we can add to NYC that will enforce a coverage
number, which might be a good idea to pull in and start incrementing as we
get progress:

https://www.npmjs.com/package/nyc#checking-coverage

I'll make those other changes this week and update the P/R!

stash:
Still just need the `.gitignore` and `.npmignore` changes

awaterma:
Updated as discussed. :)

PR #110, opened 2018-04-30
Apr2018 node versions

stash:
Support Node 10.x and 9.x

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: node_js
 node_js:
 - stable
+- "9.0"
+- "8.0"
 - "7.0"
 - "6.0"
 - "5.12"


stash:
Re-tested with `v10.2.1` locally and still works fine

BridgeAR:
@stash this fix is actually not correct. The message was never tested on older Node.js versions. The behavior is also documented: https://nodejs.org/api/assert.html#assert_assert_throws_block_error_message. The message would have been used in case __no__ error would have been thrown.

The reason why Error instances work in Node.js >= 10 supports validation objects and errors are accepted as such as well (the non-enumerable properties `message` and `name` will also be validated instead of only using all enumerable properties in case errors are passed through).

I opened https://github.com/salesforce/tough-cookie/pull/113 to address this in a backwards compatible way.

PR #113, opened 2018-05-28
Fix assert throws assertion

BridgeAR:
So far the error message was not tested when calling `assert.throws`.
In new Node.js versions it is warned in case this is detected.
Now the error message is validated by using a regular expression
instead of the string.


stash:
OK, I'll cancel #110 and merge this

PR #114, opened 2018-05-28
Fix custom inspection

BridgeAR:
From Node.js 11 on custom inspection will not work with the `inspect`
property anymore. This changes it to use the newer custom inspection
symbol if available. To stay backwarts compatible it keeps the old
function in place in case it is called directly by users.

### DIFF
--- a/lib/cookie.js
+++ b/lib/cookie.js
@@ -31,6 +31,7 @@
 'use strict';
 var net = require('net');
 var urlParse = require('url').parse;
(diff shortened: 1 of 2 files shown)


stash:
LGTM

stash:
LGTM. Will merge once the assert API incompatibility thing gets resolved

BridgeAR:
Rebased due to conflicts.

Issue #93, opened 2017-09-19
Feature request: remove all cookies from cookie jar at once


olalonde:
It would be helpful to have a method to completely clear the cookie jar.

tizmagik:
May I suggest:

```js
var tough = require('tough-cookie');
var cookiejar = new tough.CookieJar();
cookiejar.cookieMonsterNomNom();
```

ichernev:
+1

As a workaround you can switch the current jar with a brand new one, but removing all cookies is a very meaningful addition.

stash-sfdc:
This probably means adding `Store#removeAllCookies` (to pair with the existing `#removeCookies`) and then having `CookieJar#removeAllCookies`.

lewebsimple:
Can't we just delete the jar and create a new instance ?

jamiehaywood:
any movement on this FeatureRequest?

jamiehaywood:
@awaterma has this been removed from the backlog?

ShivanKaul:
@jamiehaywood  this was done in #115, going to close this issue.

PR #115, opened 2018-06-02
remove all cookies from cookie jar at once

YevhenLukomskyi:
implements feature https://github.com/salesforce/tough-cookie/issues/93

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,6 +354,14 @@ The `store` argument is optional, but must be a _synchronous_ `Store` instance i
 
 The _source_ and _destination_ must both be synchronous `Store`s. If one or both stores are asynchronous, use `.clone` instead. Recall that `MemoryCookieStore` supports both synchronous and asynchronous API calls.
 
+### `.removeAllCookies(cb(err))`
+
+Removes all cookies from the jar.
+
+### `.removeAllCookiesSync()`
+
+Sync version of `.removeAllCookies()`
+
 ## Store
 
 Base class for CookieJar stores. Available as `tough.Store`.
@@ -417,6 +425,10 @@ The implementation MUST NOT pass an error if the cookie doesn't exist; only pass
 Removes matching cookies from the store.  The `path` parameter is optional, and if missing means all paths in a domain should be removed.
 
 Pass an error ONLY if removing any existing cookies failed.
+
+### `.removeAllCookies(cb(err))`
+
+Removes all cookies from the store.
 
 ### `store.getAllCookies(cb(err, cookies))`
 
--- a/README.md
+++ b/README.md
@@ -198,7 +198,7 @@ compute the TTL relative to `now` (milliseconds).  The same precedence rules as 
 
 The "number" `Infinity` is returned for cookies without an explicit expiry and `0` is returned if the cookie is expired.  Otherwise a time-to-live in milliseconds is returned.
 
-### `.canonicalizedDoman()`
+### `.canonicalizedDomain()`
 
 ### `.cdomain()`
 
--- a/lib/cookie.js
+++ b/lib/cookie.js
@@ -1379,6 +1379,11 @@ CookieJar.prototype.clone = function(newStore, cb) {
   });
 };
 
+CAN_BE_SYNC.push('removeAllCookies');
+CookieJar.prototype.removeAllCookies = function(cb) {
+  this.store.removeAllCookies(cb);
+}
+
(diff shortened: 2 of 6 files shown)


test/api_test.js
```diff
@@ -295,6 +296,29 @@ vows
...
+
+          jar.removeAllCookiesSync();
+          
+          return jar;
+        },
+        "no cookies in the jar": function (jar) {
+          jar.store.getAllCookies(function(err, cookies) {
+            assert(err == null);
+            assert(cookies != null);
+            assert(cookies.length === 0, 'cookies were not removed')
```
stash: Typo: should be "cookies were removed"

stash:
Looks like this is a semver major kind of change since it renders older stores potentially incompatible.

test/cookie_jar_test.js
```diff
@@ -480,4 +480,29 @@ vows
...
+        return jar;
+      },
+      "no cookies in the jar": function (jar) {
+        jar.removeAllCookies(function (err) {
+          assert(err == null);
+  
+          jar.store.getAllCookies(function(err, cookies) {
+            assert(err == null);
+            assert(cookies != null);
+            assert(cookies.length === 0, 'cookies were not removed')
```
stash: Again, should say "cookies were removed"

stash:
So, thinking about this more, just to increase backwards compatibility, is there a way we can delete cookies one-by-one for old Store implementations that don't have an accelerated `removeAllCookies`?

test/api_test.js
YevhenLukomskyi: this message should specify the reason why cookies array is not empty. Isn't it?

YevhenLukomskyi:
In my opinion we could use `store.getAllCookies()` to retrieve cookies from store, and then remove them domain-by-domain using `store.removeCookies(domain)`. It will increase the complexity of implementation, but at least it will be compatible with the current version of store. @stash what do you think about this way?

stash:
@YevhenLukomskyi Yeah, I think that's appropriate. Something like

```js
if (store.removeAllCookies instanceof Function) {
  store.removeAllCookies(cb);
} else {
  store.getAllCookies(function(err, cookies) {
    // ... check error, remove one-by-one
  });
}
```

README.md
```diff
@@ -198,7 +198,7 @@ compute the TTL relative to `now` (milliseconds).  The same precedence rules as
 
 The "number" `Infinity` is returned for cookies without an explicit expiry and `0` is returned if the cookie is expired.  Otherwise a time-to-live in milliseconds is returned.
 
-### `.canonicalizedDoman()`
+### `.canonicalizedDomain()`
```
awaterma: Spelling fix. NIce!

YevhenLukomskyi:
@stash added one-by-one removal for old store implementations that don't have `removeAllCookies`

YevhenLukomskyi:
@stash @awaterma any updates on this one?

awaterma:
We want to get the versioning straight here, before merging this pull request. We still have backwards compatibility, but this is a new feature. We should probably increment the minor version, but it could be a major version kind of change. We will wait to discuss the versioning with @stash and then most probably merge before our next meeting.

@ruoho is recommending a major version change for this API modification.

lib/cookie.js
```diff
@@ -1385,6 +1385,43 @@ CookieJar.prototype.clone = function(newStore, cb) {
   });
 };
 
+CAN_BE_SYNC.push('removeAllCookies');
+CookieJar.prototype.removeAllCookies = function(cb) {
+  var store = this.store;
+  if (store.removeAllCookies instanceof Function) {
```
stash: Just noticed that this is going to evaluate to `true` for all stores that inherit from the `Store` in `lib/store.js`. This will cause the callback to receive an error instead of using the fallback code. Probably needs to be changed to:

```js
if (store.removeAllCookies instanceof Function && store.removeAllCookies != Store.prototype.removeAllCookies) {
}
```

stash: This should get tested explicitly too: create a new `Store` class via inheritance and confirm that an error isn't passed to the callback, but instead the `getAllCookies`/`removeCookie` route is taken.

stash:
@YevhenLukomskyi I'm going to accept this pull request into a branch and make some further changes. Thank you for your help!

stash:
@YevhenLukomskyi FYI this is published as v2.5.0 on npm now. Thanks for your patience.

PR #119, opened 2018-07-30
Inline version

### LINKED ISSUE #72: package.json is required thus packed by browserify
Hi,
we indirectly use your library on browser side thanks to browserify and I realized that the generated code contains your package.json (after some npm transformation).
this is due to the instruction `require('../package.json').version` in `./lib/cookie.js`
I don't see any good solution to keep the version in one place and avoiding to require the package.json unless you add some preprocessing to update package.json....

stash:
Use `genversion` to inline the version number. Automated to run during `npm version <arg>`.
Fixes #72


stash:
@ruoho ready for re-review; I moved the package-lock stuff to another PR.

PR #121, opened 2018-07-30
Upgrade to punycode 2.1

### LINKED ISSUE #82: Punycode AMD error with RequireJS
We get the following error TypeError: punycode.toASCII is not a function when using node request that uses this library, the problem is with punycode 1.4.1 which doesn't work with AMD. Please update dependencies to 2.0.1

Thank you

stash:
Fixes #82

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,6 +71,6 @@
   },
   "dependencies": {
     "psl": "^1.1.28",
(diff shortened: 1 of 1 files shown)


stash:
TravisCI test indicates that this seems to work with node >= 0.10.

PR #130, opened 2018-10-15
Fix checking function

ddo:
Not all the function is instance of Function
How did I encounter this issue: I use [got](https://github.com/sindresorhus/got) with cookieJar enable and I met the error something like ``cb is undefined``
more info: http://tobyho.com/2011/01/28/checking-types-in-javascript/


szmarczak:
@stash https://stackoverflow.com/questions/899574/what-is-the-difference-between-typeof-and-instanceof-and-when-should-one-be-used

ShivanKaul:
This should be taken care of by https://github.com/salesforce/tough-cookie/pull/178 (which also added tests)

Issue #111, opened 2018-05-09
Conflicting Paths Generated By permutePath


duper51:
I'm experiencing a bug using this library when attempting to get a previously set cookie from a JAR using the FileCookieStore() store.

If I set a certain cookie with a path ending in a slash (such as "/cas/"), when I attempt to retrieve it using getCookies using the same URL as the first parameter, the call to `permutePath` returns an Array containing `["/cas", "/"]` which doesn't match "/cas/" which causes the `pathMatcher` `matchRFC` to not find any matching cookies, even though I'm providing the same URL. I'm not entirely sure this is the desired behavior.

duper51:
Just had a look at the PRs and it looks like issue #90 is resolving this issue.

ruoho:
@duper51 Which FileCookieStore were you using?

duper51:
I'm gonna take a look at the code I was using to get this to work, I am not sure which I was using.

awaterma:
We believe this has been fixed in our 4.x release. Please reopen with a test case if there's been a mistake.

PR #131, opened 2018-10-29
Allow trailing slash in permutePath

ruoho:


test/domain_and_path_test.js
```diff
@@ -187,7 +187,7 @@ vows
       "trailing slash": {
         topic: tough.permutePath.bind(null, '/foo/bar/'),
         "three things": function (list) {
```
awaterma: Change the test title to "four things"?

ruoho: Updated

test/domain_and_path_test.js
```diff
@@ -186,8 +186,8 @@ vows
       },
       "trailing slash": {
         topic: tough.permutePath.bind(null, '/foo/bar/'),
-        "three things": function (list) {
-          assert.deepEqual(list, ['/foo/bar', '/foo', '/']);
+        "four things": function (list) {
+          assert.deepEqual(list, ['/foo/bar/', '/foo/bar', '/foo', '/']);
```
stash: Perhaps `/foo/` should be an intermediate as well? 

Overall, I think I have a performance concern: is this is going to result in a lot more hits to the underlying store?

stash:
Is this resolving #111 ?

awaterma:
We need to ensure this is correct, according to the RFC. I think that should be the determinant here. If the RFC is not clear, we will support the current behavior.

By RFC, we should support the BIS-Draft, version 3.

PR #136, opened 2018-12-31
Use ip-regex instead of net.isIP. Fix #125

### LINKED ISSUE #125: Node dependency breaks in non-node environments
Hi,
I'm working on a Nativescript project, where the api surface area is node but not quite. The library depends on the presence the node [net](https://nodejs.org/api/net.html#net_net_isip_input) api. This doesn't exist in NS, and the shim for webpack returns true for isIp and related functions [here](https://github.com/webpack/node-libs-browser/blob/1b67bdeefca87d81792e86278439213a6932170e/mock/net.js#L9). This causes [this line](https://github.com/salesforce/tough-cookie/blob/8302ebc43a11f608eb1c62715e7eb2d38920d32c/lib/cookie.js#L323) to fail. I suggest that you drop the use of the node inbuilt module and use a lib or a handwritten function to check for IPs.

gzzhanghao:
Replace [net.isIP] with [ip-regex].

[ip-regex]: https://github.com/sindresorhus/ip-regex/blob/master/index.js
[net.isIP]: https://github.com/nodejs/node/blob/master/lib/internal/net.js#L7

### DIFF
--- a/lib/cookie.js
+++ b/lib/cookie.js
@@ -29,9 +29,9 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 'use strict';
-var net = require('net');
 var urlParse = require('url').parse;
 var util = require('util');
+var ipRegex = require('ip-regex')({ exact: true });
 var pubsuffix = require('./pubsuffix-psl');
 var Store = require('./store').Store;
 var MemoryCookieStore = require('./memstore').MemoryCookieStore;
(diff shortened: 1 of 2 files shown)


stash:
Looks good to me. Breaks compatibility with node 0.10 and 0.12 though.

stash:
`request` now has `"node": ">= 6"` and `jsdom` has `">=8" so maybe we should roll forward support to at least node 6?

Turbo87:
@stash this has broken Node 4 and 6 compat too. `ip-regex@3` is Node 8+, but `tough-cookie` (according to `package.json`) supports Node 6+. ultimately this is causing Jest to no longer work on Node 6 when installed via `yarn`.

```
=> Found "ip-regex@3.0.0"
info Reasons this module exists
   - "jest#jest-cli#jest-environment-jsdom#jsdom#request-promise-native#tough-cookie" depends on it
```

PR #146, opened 2019-02-05
.cloneSync should not require object param

AlexOwl:
Documentation says: `.cloneSync([store])` 
So cloneSync does not require `store` param, but it does due to property(`synchronous`) checking


lib/cookie.js
```diff
@@ -1436,6 +1436,8 @@ CookieJar.prototype.removeAllCookies = function(cb) {
 
 CookieJar.prototype._cloneSync = syncWrap('clone');
 CookieJar.prototype.cloneSync = function(newStore) {
+  if(arguments.length === 0)
```
olleolleolle: Two syntactic details which would make this code change blend in even better: 

- even though this `if` has a 1-line body, use the `{` and `}` around it, just like the `if` just after this
- add a space after the `if` word – so that `if (arguments.length === 0) {`

AlexOwl: Recheck it

AlexOwl:
Any update on this?

medelibero-sfdc:
@AlexOwl Sorry for the radio silence on this. Is there any way you could resolve the `lib/cookie.js` conflicts? Once that happens I will review and hopefully get it merged in.

Issue #150, opened 2019-02-26
ip-regex dependency breaks on old very old node versions


Ankcorn:
Hi, 

I'm currently working on a project using a specific (very old) node version. 

ip-regex's use of es6 features makes our application crash on startup  with this new update and because we have no control over the library pulling in tough cookie we cannot backport to version 2.5 easily and might have to deal with breakages in the future. 

Would it be possible to remove ip-regex as a dependency?

Thanks
Tom

amol-c:
Facing the same issue here.

awaterma:
Ip-regex has been removed as a dependency:

https://github.com/salesforce/tough-cookie/blob/master/package.json#L104

PR #151, opened 2019-02-26
Removed ip-regex dependency and added the code locally including tests

ThomasAnkcorn:
Fixes issue #150

Info for tests was found https://en.wikipedia.org/wiki/IP_address

and ipv6 regex is based on the work here https://gist.github.com/syzdek/6086792


stash:
Hi Thomas! Thanks for submitting this pull request. Coincidentally, I've actually started on the same thing in another branch and hadn't pushed it yet! It is now pushed in #152 .  @ThomasAnkcorn, would you be interested in reviewing my work?

If we do go with this pull request, the license headers have an incorrect copyright year. The tests look pretty good, but I'd like to see the integration test from this PR as well.

ThomasAnkcorn:
@stash Thanks for making the change! I can make a pull request into your branch with the tests from this and the ip-regex module. Then this pull request could be closed.

stash:
@ThomasAnkcorn sure, that would work. I'll close this one.

PR #152, opened 2019-02-27
Embed IP address regex and thoroughly test

stash:
Removes the dependency on `ip-regex` and adds test coverage on how exactly tough-cookie is treating IP addresses.


stash:
See also: #151

ruoho-sfdc:
@stash Is this the IPv6 pull request you mentioned today?

PR #158, opened 2019-05-30
fix(cookie-jar): use typeof for checking type of option

SimenB:
Works around https://github.com/facebook/jest/issues/2549 when used in Jest

### DIFF
--- a/lib/cookie.js
+++ b/lib/cookie.js
@@ -995,7 +995,7 @@ CAN_BE_SYNC.push('setCookie');
 CookieJar.prototype.setCookie = function(cookie, url, options, cb) {
   var err;
   var context = getCookieContext(url);
-  if (options instanceof Function) {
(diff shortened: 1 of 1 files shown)


ianbytchek:
Ping! Can this be merged now?

alvis:
is there anything blocking this from merging?

ShivanKaul:
LGTM to me too. Shipping 🚢

theKashey:
Any ETA to actually release at least as a beta?

awaterma:
We'll get a release discussion going in our next meeting. We have a few more p/rs we want to bring in, I think.

I think you can do some local testing with our master branch by pulling in a GitHub dependency into your package.json.

theKashey:
Went with [patch-package](https://github.com/ds300/patch-package) solution for now.

PR #164, opened 2019-07-03
Add support for cookie prefixes

### LINKED ISSUE #81: draft-ietf-httpbis-cookie-prefixes
https://tools.ietf.org/html/draft-ietf-httpbis-cookie-prefixes-00

>    Section 8.5 and Section 8.6 of [RFC6265] spell out some of the
>    drawbacks of cookies' implementation: due to historical accident, it
>    is impossible for a server to have confidence that a cookie set in a
>    secure way (e.g., as a domain cookie with the "Secure" (and possibly
>    "HttpOnly") flags set) remains intact and untouched by non-secure
>    subdomains.
> 
>    We can't alter the syntax of the "Cookie" request header, as that
>    would likely break a number of implementations.  This rules out
>    sending a cookie's flags along with the cookie directly, but we can
>    smuggle information along with the cookie if we reserve certain name
>    prefixes for cookies with certain properties.
> 
>    This document describes such a scheme, which enables servers to set
>    cookies which conforming user agents will ensure are "Secure", and
>    locked to a domain.

Parser should respect cookies starting with `__Secure-` and `__Host-`.

(Probably use `indexOf` in the implementation; possible ReDoS)

ShivanKaul:
- Waiting on #128 to get merged so can use README blurb about draft rfc6265bis
- Would like to get feedback on how to resolve TODO at https://github.com/ShivanKaul/tough-cookie/blob/1fe14103d068edf1536e1e5e4ec85c97db37e3be/lib/cookie.js#L1121

### DIFF
--- a/lib/cookie.js
+++ b/lib/cookie.js
@@ -88,9 +88,9 @@ function checkSameSiteContext(value) {
 }
 
 const PrefixSecurityEnum = Object.freeze({
-  SILENT: 'silent',
-  STRICT: 'strict',
-  DISABLED: 'unsafe-disabled'
+  SILENT: "silent",
+  STRICT: "strict",
+  DISABLED: "unsafe-disabled"
 });
 
 /*
@@ -602,11 +602,11 @@ function parse(str, options) {
  *  If the cookie-name begins with a case-sensitive match for the
  *  string "__Secure-", abort these steps and ignore the cookie
  *  entirely unless the cookie's secure-only-flag is true.
- * @param c
+ * @param cookie
  * @returns boolean
  */
-function securePrefixConditionMet(disabled, c) {
-  return disabled || !c.key.startsWith("__Secure-") || c.secure;
+function isSecurePrefixConditionMet(cookie) {
+  return !cookie.key.startsWith("__Secure-") || cookie.secure;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


awaterma:
could you be sure that this p/r is based on pull request #155? That should make merging easier to do. :)

ShivanKaul:
Yes, once that one is merged in I shall rebase this one.

lib/cookie.js
```diff
@@ -980,15 +1006,36 @@ function CookieJar(store, options) {
   if (options.looseMode != null) {
     this.enableLooseMode = options.looseMode;
   }
+  if (options.prefixSecurity != null) {
+    const normalizedPrefixSecurity = options.prefixSecurity.toLowerCase();
+    switch (normalizedPrefixSecurity) {
+      /* The three supported options */
+      case PrefixSecurityEnum.STRICT:
+      case PrefixSecurityEnum.SILENT:
+      case PrefixSecurityEnum.DISABLED:
```
awaterma: Why do we need the multi-line CASE here? Future proofing? Just wondering if this might be better expressed for now as a simple setter.

lib/cookie.js
```diff
@@ -1070,6 +1117,24 @@ CookieJar.prototype.setCookie = function(cookie, url, options, cb) {
     return cb(options.ignoreError ? null : err);
   }
 
+  /* 6265bis-02 S5.4 Steps 15 & 16 */
+  /*  TODO: unsure whether this ignore error condition should check for options.ignoreError
+      || or && with this.prefixSecurity?
+   */
+  const ignoreErrorForPrefixSecurity = this.prefixSecurity === PrefixSecurityEnum.SILENT;
```
awaterma: Any suggestions on how to confirm what the right behavior should be here? I'll try and review the RFC on this point, unless you have a better idea. :)

awaterma: Perhaps we could augment the unit test around this as well; I'd like to see a case for PrefixSecurityEnum.SILENT.

lib/cookie.js
ShivanKaul: Because we are only accepting those 3 options, right? If it's not one of the 3 options, then we opt for the default (silent).

Issue #144, opened 2019-01-31
Cannot read property secure of undefined


nonyy:
Sample cookie I am trying to parse is included. Any tips on parsing?
Code:
https://repl.it/repls/ProfitableUnfortunateFossil

aranmohyeddin:
@awaterma august has ended. I also have this issue.

awaterma:
I've created a unit test for this now, working on finding time to resolve the issue.

awaterma:
Could you two try this with the master branch? I can't reproduce this:

https://github.com/salesforce/tough-cookie/compare/master...awaterma:issue%23144_property_secure

aranmohyeddin:
I ended up changing my whole authentication system from session auth to JWT auth because I couldn't make cookies and csrf work:D

awaterma:
This is not a bug. You can't call a method that takes a callback without a callback.

I wonder though, should we make this an API error? To clarify for users?

@medelibero-sfdc

awaterma:
@medelibero-sfdc -- let me know what you think, we can open another issue for error handling in async if we want to. I've got a unit test that proves this out -- that's the latest p/r.

medelibero-sfdc:
@awaterma I agree we should add in some extra error handling for this case so people get at least a more friendly exception. But I agree that this specific issue can be closed out and to create a new issue for the error handling bit.

awaterma:
Let's discuss in our next meeting. I think we should do the same as NodeJS when callers execute async methods incorrectly; which may very well be nothing. Let's investigate and create a story for how we want things to be, and then prioritize. :)

PR #170, opened 2019-09-22
Added unit test for issue 144; passing in master.

awaterma:
Additional unit test for #144.

### DIFF
--- a/test/cookie_jar_test.js
+++ b/test/cookie_jar_test.js
@@ -644,4 +644,30 @@ vows
       }
     }
   })
+  .addBatch({
+    "Issue #144 - secure": {
+      "of undefined": {
+        topic: function() {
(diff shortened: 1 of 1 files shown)


awaterma:
@ShivanKaul -- would you mind taking a look? I think this is a simple p/r.

awaterma:
I'll sign these and update.

medelibero-sfdc:
@awaterma Your test uses `getCookiesSync` while the problem referenced in issue 144 uses `getCookies`. If you use `getCookies` you will get the error but if you use `getCookiesSync` it works.

awaterma:
Thanks @medelibero-sfdc! I've modified the test and reproduced locally. I'll see if I can fix the underlying issue. I'll close this out and resubmit later.

PR #172, opened 2019-10-05
Issue 160 local suffix

### LINKED ISSUE #160: Allow .local suffix domains for testing purposes
Hi,

I am using request (which uses tough-cookie) to test a web app that is hosted on a domain ending with a .local suffix (ex: http://foo.bar.local). Once the app's server receives a GET request, it sends back a redirect status with a set-cookie header along with authentication cookies. Each containing a .bar.local domain. When the client responds to the redirect, it attempts to put the stored cookies in the response header. This fails as the permuteDomain function (permuteDomain.js) calls the psl.get(domain) method in order to obtain the public suffix of the current page (in this case foo.bar.local). This method returns null when it is passed a domain ending in .local. This causes permuteDomain to set the list of permutations to contain foo.bar.local as the only permutation in this list. Consequently, the cookies fail to be added to the response header in the MemoryCookieStore.findCookies method.

Testing on a .local domain is fairly commonplace and, correct me if I am wrong, but .local is considered public suffix. If the rejectPublicSuffixes option is initially set to false, shouldn't .local domains be allowed as well? This boolean could be passed down to findCookies and permuteDomain, allowing the function to find the permutations of domains ending with a .local suffix. I would be happy to implement this and submit a PR.

miggs125:
fixes #160 

I have added an additional CookieStore option named allowSpecialUseDomain to allow for [special-use domains](https://www.iana.org/assignments/special-use-domain-names/special-use-domain-names.xhtml) such as '.local'. This is useful for testing purposes, '.local' is a commonly used suffix in testing environments.The const array in permuteDomain includes only .local because that is what I mentioned in the original issue. However, I could add the rest if needed.

Here are some more links regarding special-use domains
https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains#Special-Use_Domains
https://tools.ietf.org/html/rfc6761
passed all tests. Ran tests with ```npm run test``` 
Updated the README to reflect changes made


miggs125:
Closed because I was having issues with linking commits to github email. will create a new PR.

PR #177, opened 2019-12-03
Re-add cookie prefix feature

ShivanKaul:
Refer to https://github.com/salesforce/tough-cookie/pull/164

### DIFF
--- a/lib/cookie.js
+++ b/lib/cookie.js
@@ -87,6 +87,12 @@ function checkSameSiteContext(value) {
   }
 }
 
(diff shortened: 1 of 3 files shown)


lib/cookie.js
```diff
@@ -87,6 +87,12 @@ function checkSameSiteContext(value) {
   }
 }
 
+const PrefixSecurityEnum = Object.freeze({
```
awaterma: That's a very nice way to declare an immutable object. :)

PR #181, opened 2020-01-02
Awaterma issue#144 property secure

awaterma:
Simple test for issue #144 that proves it is not an issue in the codebase. Update Dockerfile to explicitly use Node 12 LTR.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node
+FROM node:12
 MAINTAINER awaterman@salesforce.com
 LABEL Description="Vendor=\"Salesforce.com\" Version=\"1.0\""
 RUN apt-get update && \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node
+FROM node:12
```
medelibero-sfdc: Is there a reason we went with version 12 instead of `latest` or `13`?

awaterma: 13 isn't a LTS (long term supported release). Latest is a bit ephemeral; it doesn't always point to the latest build. More here: https://medium.com/@mccode/the-misunderstood-docker-tag-latest-af3babfd6375

Here's some info on Node releases and LTS: https://nodejs.org/en/about/releases/

medelibero-sfdc: Ahh got it. Fair enough. I just wanted to make sure.

PR #185, opened 2020-02-27
Remove ip-regex dependency

ruoho:


awaterma:
@ruoho -- should this be closed?

PR #189, opened 2020-03-19
Create CHANGELOG.md

ShivanKaul:
Better late than never 😄

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,12 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+## 4.0.0
+
+- SameSite cookie support
+- Cookie prefix support 
+- Support for promises
+- Use ESLint and Prettier to apply consistent, modern formatting
+- '.local' support 
+- Numerous bug fixes!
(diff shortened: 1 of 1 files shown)


grantila:
Great! But it would be nice to know why major, i.e. what are the breaking changes.

CHANGELOG.md
```diff
@@ -0,0 +1,12 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+## 4.0.0
+
```
awaterma:
```suggestion
### Breaking Changes (Major Version)
- Modernized JS Syntax
  - Use ESLint and Prettier to apply consistent, modern formatting
- Support for Promises
```

CHANGELOG.md
```diff
@@ -0,0 +1,12 @@
...
+- SameSite cookie support
+- Cookie prefix support 
+- Support for promises
+- Use ESLint and Prettier to apply consistent, modern formatting
+- '.local' support 
+- Numerous bug fixes!
```
awaterma:
```suggestion
### Minor Changes
- SameSite cookie support
- Cookie prefix support 
- '.local' support 
- Numerous bug fixes!
```

awaterma:
@ShivanKaul -- let me know what you think; perhaps there's an even better way to setup our changelog. 

I wonder if we should also consider adding something like this as a Githook?

PR #196, opened 2020-06-17
Returned null in case of empty cookie value

vsin12:
1. some of the old sites by mistake send a blank cookie, and whenever a blank cookie is encountered, failed to parse cookie error comes up - so I made a fix of checking the length and accordingly calling callback with null value.

Note: Either this fix, or CookieJar constructor needs to take SetCookieOptions too, as ignoreError is part of SetCookieOptions interface

### DIFF
--- a/lib/cookie.js
+++ b/lib/cookie.js
@@ -1081,6 +1081,9 @@ class CookieJar {
     if (typeof options === "function") {
       cb = options;
       options = {};
+
+      if(typeof cookie === 'string' && cookie.length <= 0 )
+        return cb(null);
     }
 
     const host = canonicalDomain(context.hostname);
--- a/lib/cookie.js
+++ b/lib/cookie.js
@@ -1099,10 +1099,12 @@ class CookieJar {
     if (validators.isFunction(options)) {
       cb = options;
       options = {};
-
-      if(!validators.isNonEmptyString(cookie))
-        return cb(null);
     }
+
+    if(!validators.isNonEmptyString(cookie) && !validators.isObject(cookie)) {
+      return cb(null);
+    }
+
     validators.validate(validators.isFunction(cb), cb);
 
     const host = canonicalDomain(context.hostname);
--- a/lib/cookie.js
+++ b/lib/cookie.js
@@ -1100,7 +1100,7 @@ class CookieJar {
       cb = options;
       options = {};
 
-      if(typeof cookie === 'string' && cookie.length <= 0 )
+      if(!validators.isNonEmptyString(cookie))
         return cb(null);
     }
     validators.validate(validators.isFunction(cb), cb);
--- a/lib/cookie.js
+++ b/lib/cookie.js
@@ -1101,11 +1101,11 @@ class CookieJar {
       options = {};
     }
 
-    if(!validators.isNonEmptyString(cookie) && !validators.isObject(cookie)) {
+    validators.validate(validators.isFunction(cb), cb);
+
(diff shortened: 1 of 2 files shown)


ruoho:
Hi @vsin12, would you be able to update the PR with a test?

vsin12:
@ruoho : Sorry, but I didn't get you ? Do you want me to give a scenario while it was breaking or you want me to write a test to check my pr ?

awaterma (changes requested):
You wrote: `Note: Either this fix, or CookieJar constructor needs to take SetCookieOptions too, as ignoreError is part of SetCookieOptions interface`

ignoreError doesn't apply to storage: `ignoreError - boolean - default false - silently ignore things like parse errors and invalid domains. Store errors aren't ignored by this option` -- so I'm not sure we need to do any more here, in terms of implementation. We might want to start a project to review ignoreError and its applications --- it might be better to have that apply globally, but we'd have to think that through.

**In terms of this P/R; I think I'd like to see the change use validators, as suggested, and then have a simple unit test added that demonstrates the use-case (an empty cookie).**

lib/cookie.js
```diff
@@ -1099,6 +1099,9 @@ class CookieJar {
     if (validators.isFunction(options)) {
       cb = options;
       options = {};
+
+      if(typeof cookie === 'string' && cookie.length <= 0 )
+        return cb(null);
     }
     validators.validate(validators.isFunction(cb), cb);
```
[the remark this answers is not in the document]
awaterma: Meaning line 1103/1106 `validators.validate(validators.isFunction(cb), cb);`

awaterma:
I think we still need a unit test for this fix. :)

medelibero-sfdc:
@vsin12 can you change this PR to allow edits from maintainers? I have a few changes I would like to push to this PR.

vsin12:
> @vsin12 can you change this PR to allow edits from maintainers? I have a few changes I would like to push to this PR.

Done.

lib/cookie.js
```diff
@@ -1100,8 +1100,13 @@ class CookieJar {
       cb = options;
       options = {};
     }
+
     validators.validate(validators.isFunction(cb), cb);
 
+    if(!validators.isNonEmptyString(cookie) && !validators.isObject(cookie) && ( cookie instanceof String && cookie.length == 0)) {
+      return cb(null);
+    }
+
```
awaterma, line 1109: Just looked at this in context, and I think this is fine. Will discuss today.

test/cookie_jar_test.js
```diff
@@ -699,4 +699,21 @@ vows
...
+      "with missing parameters": {
+        topic: function() {
+          const jar = new tough.CookieJar();
+          jar.setCookie(
+            '',
+            "https://google.com",
+            this.callback
+          );
+        },
+        "results in a error being returned because of missing parameters": function(err, cookies) {
```
awaterma, line 713: We don't actually return an error, we are invoking the callback with a Null here. Should we invoke the CB with an Error()?

awaterma (commented):
I'll discuss this today; I think we want to add a suggestion to return an error when presented with an empty cookie.

lib/cookie.js
awaterma: Please see comment below; I think we may want an Error here.

awaterma (approved):
Approved. We may want to add error checking/throwing at some point, but that's probably in a major release.

PR #203, opened 2020-09-20
Add explanation and more special-use domains

### LINKED ISSUE #174: Fully implement RFC 6761
We may need to do more to fully support RFC 6761:

https://tools.ietf.org/html/rfc6761

This issue is to mark a Spike on that for the new 4.x release.

ShivanKaul:
Fix #174

I think @ruoho said he'd write a test case for #173 - once that is in I'll rebase off of that and add some more tests.

The hack in https://github.com/salesforce/tough-cookie/blob/master/lib/permuteDomain.js#L41 was introduced in #173 and is expanded in this PR because we rely on PSL's `psl.get()` call. This doesn't work for special-use domains because most of them are not in the Public Suffix List (except for `.onion`). Interestingly, `libpsl` (the C equivalent of the PSL js library we use) has a `isCookieDomainAcceptable()` function that would be useful - I raised an issue asking if there's interest in it for the PSL js library here: https://github.com/lupomontero/psl/issues/232

### DIFF
--- a/lib/permuteDomain.js
+++ b/lib/permuteDomain.js
@@ -33,15 +33,27 @@ const pubsuffix = require("./pubsuffix-psl");
 
 // Gives the permutation of all possible domainMatch()es of a given domain. The
 // array is in shortest-to-longest order.  Handy for indexing.
-const SPECIAL_USE_DOMAINS = ["local"]; // RFC 6761
+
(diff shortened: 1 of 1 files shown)


medelibero-sfdc (approved):
lgtm

PR #207, opened 2020-10-23
Adds travis jobs on ppc64le

dineshks1:
This PR adds power support on travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+arch:
+ - amd64
+ - ppc64le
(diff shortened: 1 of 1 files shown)


awaterma:
@dineshks1 -- could you please sign the salesforce-cla? We should then be able to accept this travis update.

awaterma:
We don't believe this is necessary, as the NodeJS project should be testing binary compatibility on other platforms. Our code is just JS after all ...

Issue #213, opened 2021-05-05
Migrate from Travis to GitHub Actions; actions implementation needs to be kept in sync with NodeJS releases ...


awaterma:
We need to keep our Travis testing in sycn with NodeJS releases: https://nodejs.org/en/about/releases/

awaterma:
We want to also add another integration testing framework that won't limit our use, but we think we want to retain Travis.

awaterma:
Looking at: https://github.com/actions/setup-node, for this work.

PR #218, opened 2021-09-21
Issue #213 -- added gh-actions flow for building and testing tough-co…

awaterma:
…okie with all current version of node. This should replace our Travis testing.


awaterma:
@ruoho-sfdc -- I think we'll have to ignore travis issues until we can get this change in; I wonder if I should also remove the .travis file? 

Anyway, simple change based on NodeJS Github Actions template: https://docs.github.com/en/actions/guides/building-and-testing-nodejs-or-python

I've tested actions with a simple private repository, I think this will work well for us.

ruoho-sfdc (approved):

PR #220, opened 2021-09-28
Issue #210 -- Updated workflow to use npm install.

awaterma:
Have to make a change to the actions commands; `npm ci` only works if an install has already happened. :)


medelibero-sfdc (approved):
LGTM

PR #223, opened 2021-10-26
Adding Updating CODEOWNERS with ECCN as per Export Control Compliance

svc-scm:


CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+# Comment line immediately above ownership line is reserved for related gus information. Please be careful while editing.
```
awaterma, line 1:
```suggestion
# Comment line immediately above ownership line is reserved for related information. Please be careful while editing.
```

awaterma: @wjhsf @ruoho -- I think we should just update the p/r with this change that removes GUS, and then accept. Does that sound good to you two?

colincasey (dismissed):

wjhsf (dismissed):

awaterma (approved):

awaterma:
@colincasey @ruoho @wjhsf -- can someone please take a look and approve? I think this file is setup as we would like now; just need another approval to merge.

wjhsf (approved):

PR #225, opened 2021-11-14
Unit test cases for `allowSpecialUseDomain` option

### LINKED ISSUE #206: Write test cases for PR 173

colincasey:
This PR adds test cases for the `allowSpecialUseDomain` option introduced in [PR #173](https://github.com/salesforce/tough-cookie/pull/173).  The special use domain check was changed to be a refinement on the public suffix check since both setting and retrieving cookies can fail since not all special use domains are considered public suffixes.

The following cases are tested:
* when the cookie jar is configured with `allowSpecialUseDomain=true`, the code will skip the lookup of a public suffix when a special use domain is detected
* when the cookie jar is configured with `allowSpecialUseDomain=false`, the code will raise an error describing proper usage when a special use domain is detected

Fixes #206

### DIFF
--- a/lib/cookie.js
+++ b/lib/cookie.js
@@ -1166,7 +1166,10 @@ class CookieJar {
 
     // S5.3 step 5: public suffixes
     if (this.rejectPublicSuffixes && cookie.domain) {
-      const suffix = pubsuffix.getPublicSuffix(cookie.cdomain());
+      const suffix = pubsuffix.getPublicSuffix(cookie.cdomain(), {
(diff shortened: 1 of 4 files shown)


wjhsf (approved):

awaterma:
I should have some time to review this today; sorry for the delay! :)

awaterma:
I've merged in our latest changes; should be able to take a deeper look at the implementation this afternoon.

awaterma (approved):
LGTM; nice changes, I think a bit clearer now.

PR #230, opened 2022-01-23
docs: update synopsis example to async await

jimmywarting:


wjhsf:
Thanks for the contribution! Closing this in favor of #231, which will include this as part of a rewrite of the entire readme.

PR #231, opened 2022-03-02
Modernize README.

wjhsf:
Focused on highlighting async methods over callbacks, while trying not to completely rewrite the doc. Also rewrote random other bits and pieces here and there.


awaterma (changes requested):
Would be great to rework the statement about AVA and the versions of NodeJS we support; besides that it looks very good to me, and a nice update. Thank you for spending time on this! Greatly appreciated.

wjhsf:
Closed in favor of #234.

PR #237, opened 2022-05-02
Stop using the internal NodeJS punycode module

gboer:
The way the code is right now, it doesn't use the Punycode package, but the internal NodeJS punycode module. That is deprecated now, which is why I saw the warning popping up (for the ones googling this, the warning is "DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead"). According to the docs of the Punycode package, you need to use `require('punycode/')`, to use the Punycode package rather than the internal NodeJS module, see https://github.com/mathiasbynens/punycode.js#installation.

### DIFF
--- a/lib/cookie.js
+++ b/lib/cookie.js
@@ -29,7 +29,7 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 "use strict";
-const punycode = require("punycode");
+const punycode = require("punycode/");
 const urlParse = require("url-parse");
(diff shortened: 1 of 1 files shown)


gboer:
Signed the license agreement, but the process is a bit wonky with some errors thrown here and there. Hope it automatically updates itself after some time?

gboer:
I'll open a new one. No idea what's wrong with the CLA process.

PR #240, opened 2022-06-23
fix: distinguish between no samesite and samesite=none

### LINKED ISSUE #191: Impossible to distinguish between unspecified SameSite and SameSite=None
Currently, both of these calls will result in a cookie with `sameSite: 'none'`:

```
> tough.parse('foo=bar').sameSite
'none'
> tough.parse('foo=bar; SameSite=None').sameSite
'none'
```

This means we can't distinguish between `SameSite=None` and no SameSite declaration at all. 

This causes problems when attempting to set a default `SameSite` that is *not* 'none'. For example, when trying to emulate the behavior of Chrome and Firefox (both of which are moving towards `SameSite=Lax` as the default for an unspecified `SameSite` attribute), it's impossible to tell if the default behavior should be applied, or if the `Set-Cookie` explicitly passed `None`, because both result in `sameSite: 'none'`.

For now, I'm using a workaround like this to determine if SameSite is really unspecified or if it's explicitly `None`:

```js
const sameSiteNoneRe = /; +samesite=(?:'none'|"none"|none)/i
const cookie = tough.parse(setCookieHeader)
const hasUnspecifiedSameSite = cookie.sameSite === 'none' && !sameSiteNoneRe.test(setCookieHeader)
```

colincasey:
closes #191

### DIFF
--- a/lib/cookie.js
+++ b/lib/cookie.js
@@ -619,11 +619,11 @@ function parse(str, options) {
           case "lax":
             c.sameSite = "lax";
             break;
+          case "none":
+            c.sameSite = "none";
+            break;
           default:
-            // RFC6265bis-02 S5.3.7 step 1:
-            // "If cookie-av's attribute-value is not a case-insensitive match
-            //  for "Strict" or "Lax", ignore the "cookie-av"."
-            // This effectively sets it to 'none' from the prototype.
+            c.sameSite = undefined;
             break;
         }
         break;
@@ -807,7 +807,7 @@ const cookieDefaults = {
   pathIsDefault: null,
   creation: null,
   lastAccessed: null,
-  sameSite: "none"
+  sameSite: undefined
 };
 
 class Cookie {
@@ -1221,7 +1221,11 @@ class CookieJar {
     }
 
     // 6252bis-02 S5.4 Step 13 & 14:
-    if (cookie.sameSite !== "none" && sameSiteContext) {
+    if (
+      cookie.sameSite !== "none" &&
+      cookie.sameSite !== undefined &&
+      sameSiteContext
+    ) {
       // "If the cookie's "same-site-flag" is not "None", and the cookie
       //  is being set from a context whose "site for cookies" is not an
       //  exact match for request-uri's host's registered domain, then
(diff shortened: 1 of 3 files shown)


lib/cookie.js
```diff
@@ -619,11 +619,11 @@ function parse(str, options) {
...
             break;
+          case "none":
+            c.sameSite = "none";
+            break;
           default:
-            // RFC6265bis-02 S5.3.7 step 1:
-            // "If cookie-av's attribute-value is not a case-insensitive match
-            //  for "Strict" or "Lax", ignore the "cookie-av"."
-            // This effectively sets it to 'none' from the prototype.
+            c.sameSite = undefined;
```
wjhsf, line 626: How does behavior change with `sameSite` explicitly set to `undefined` vs unset (implicitly `undefined`)?

colincasey: it would catch cases where the SameSite attribute is defined but the value is invalid.  e.g.; `a=0; SameSite=bad`

lib/cookie.js
```diff
@@ -807,7 +807,7 @@ const cookieDefaults = {
   pathIsDefault: null,
   creation: null,
   lastAccessed: null,
-  sameSite: "none"
+  sameSite: undefined
```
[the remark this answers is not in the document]
colincasey: yeah, i'm torn on this detail.  we could change it to `null` to be more consistent but `undefined` seems to convey the intent better.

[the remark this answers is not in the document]
awaterma: As we discussed in our meeting; I think `undefined` is a good choice. This is a bit hard to do, since we really built this as part of an earlier draft. But inline with that; I think `undefined` works well.

awaterma (approved):
I think we should add a note for this p/r into the changelog for 4.1; we can do that in a separate p/r, but I think we may want to do it here so we don't miss the note on release.

awaterma:
@wjhsf -- anything holding us back here? Perhaps we can try and discuss in our meeting tomorrow?

wjhsf (approved):

lib/cookie.js
awaterma: @wjhsf -- can you please let me know if you think we can resolve this issue?

PR #241, opened 2022-06-23
fix: cache psl lookup

### LINKED ISSUE #239: expensive blocking PSL lookup
https://github.com/salesforce/tough-cookie/blob/9b101310dd73f55b71e8120f50b9629e91ee0121/lib/pubsuffix-psl.js#L67

at high throughputs, PSL lookups can be quite intensive. for example, with 1000 calls to psl.get, this can take 400 milliseconds, and completely blocks the event loop. this seems to relatively linearly scale with more calls.

a recommendation is to use a simple Map cache; for example, 

`return psl.get(domain);` 

would be transformed to

```
if (!cache.has(domain)) {
   cache.set(domain, psl.get(domain));
}

return cache.get(domain);
```

(with cache initialized at the top as a new Map).

As the psl.get results are static, this would only benefit the performance, with no drawbacks.

colincasey:
closes #239

### DIFF
--- a/lib/pubsuffix-psl.js
+++ b/lib/pubsuffix-psl.js
@@ -40,6 +40,8 @@ const SPECIAL_USE_DOMAINS = [
   "test"
 ];
 
+const cache = new Map();
+
 function getPublicSuffix(domain, options = {}) {
   const domainParts = domain.split(".");
   const topLevelDomain = domainParts[domainParts.length - 1];
@@ -64,7 +66,11 @@ function getPublicSuffix(domain, options = {}) {
     );
   }
 
-  return psl.get(domain);
+  if (!cache.has(domain)) {
+    cache.set(domain, psl.get(domain));
+  }
+
+  return cache.get(domain);
 }
 
 exports.getPublicSuffix = getPublicSuffix;


wjhsf (changes requested):
This is kind of not our issue - if `psl` can improve its performance, then caching might not even be necessary. Alternatively, are there other packages that could provide better performance than `psl`?

If we do decide to cache lookups, it will need be a proper cache in order to avoid a spike in memory usage. Using just a `Map` is not sufficient, as it has no eviction mechanism. To achieve that, we have a few options:
1. Provide a cache, but do not expose any configuration options
2. Provide a cache and expose configuration options
3. Support a cache, but require users to provide their own

There are pros and cons to all three. The first is minimal work and does not change our interface, but users might not like whatever caching behavior we choose. The alternatives give more control to the users, but add complexity. Particularly, the cache is a global state for what is mostly an implementation detail, and I don't know that we have any established patterns for something like that.

Having thought a bit about this (but probably not enough), my current best idea is adding a `_cache` property to `getPublicSuffix`. That makes it clear that it's primarily intended for internal use, but users who want finer control can replace it with their own implementation.

gboer:
It seems someone else already found the same issue and has a fix open for it: https://github.com/lupomontero/psl/issues/301.

remusao:
Hi, in case there is interest for an alternative library that is built around performance[1] (without relying on any caching; hence better memory usage) maybe [`tldts`](https://github.com/remusao/tldts/) would be of interest as a replacement for `tough-cookie`. I took the liberty to check what it would take to migrate from `psl` and it's pretty straight forward: https://github.com/remusao/tough-cookie/commit/5439da617b2ac91001b3626d8a4ca7f0dda81bfe

Disclaimer: I am the maintainer of `tldts`.

[1] Some benchmarks supporting the performance claims can be found here: https://remusao.github.io/posts/tldts-benchmarks.html (~670x faster than `psl` at the time of writing).

colincasey:
as for this PR, i'm closing it and we'll (hopefully) sort out these performance issues in our next release

Issue #246, opened 2022-08-22
4.1 introduced breaking changes


Twipped:
Our jest environment just started throwing errors because `allowSpecialUseDomain` is now being more enforced in this new release and `jest-environment-jsdom` doesn't set it. I tried to correct for it via Jest's `testEnvironmentOptions` config passthru, but that produced other problems. You may want to revert that release, it's gonna break a WHOLE LOT of builds.

gbettencourt:
Also breaking our jest tests.  Don't see any breaking changes mentioned in the release notes.

arpowers:
Same here! Happy that this isn't a bug on our end, was about to throw the computer out the window.

arpowers:
@awaterma after reviewing a few things, this is definitely a breaking change and probably causing lots some headaches right now.  Recommend publishing a rollback minor release and pushing v5 with any updates.  For now, we are manually overriding the dep via PNPM as we can't update JSDom ourselves and there provide no easy way of patching this (e.g. to provide the options recommended in the error)...

Jhzl2:
Our jest environment also failed in our tests. We work on a react app but we don't use tough-cookie. We assume that it is a dependency of some installed library. Could you give me a hand?
<img width="598" alt="image" src="https://user-images.githubusercontent.com/101297509/186033688-6d70ea65-dee4-417b-afa5-bd02ddc20cd9.png">

awaterma:
I'm sorry this is causing you all some issues! For some background on how to use localhost as a special use domain, please see the discussion and tests in issue #215.

I believe this is coming up for v4.1.0 as @ShivanKaul implemented all of the "special-use" domains per RFC 6761 in the following p/r: https://github.com/salesforce/tough-cookie/pull/203/files, which we included in this release. We view this as continued work for the RFC that we started in v4.0.0.

In regards to working with cookies in `localhost` please see: https://www.rfc-editor.org/rfc/rfc6761.html#section-6.3

I will reach out to the team in Slack and see if we can get some discussion going. Sorry to cause issues for those that depend upon us; we are working to improve `tough-cookie` and keep up to date with the relevant RFCs to support cookies correctly into the future!

One direction we could go would be to set `rejectPublicSuffixes` to `false` in our default behavior to work around these issues for the `4.x.y` line.

@colincasey

arpowers:
Most of us can certainly empathize with these type of challenges so no worries there. We appreciate the work on keeping things compliant. 

[...]

PR #253, opened 2022-08-25
fix: allow set cookies with localhost

colincasey:
Adding more tests to cover the breaking use cases noted in #246. e.g.; `new CookieJar().setCookieSync("settingThisShouldPass=true; Domain=localhost; Path=/;", "http://localhost")`

Also:
* modifies the assertion for a test introduced in #221 that appears to be incorrect
* manually validated using a default jest environment configured for `jsdom` which only allows cookies to be set after applying this fix


awaterma:
Thanks! I'm happy with this approach! Let's discuss a bit and merge, and then generate another patch release for our week of excitement. :)

awaterma (approved):
Approved; will merge after discussion.

awaterma (approved):
Approved

PR #258, opened 2022-12-19
Update universalify dependency

stianjensen:
To a more modern version:
https://github.com/RyanZim/universalify/releases

This will update the minimum supported node version to 10.x

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,7 @@
     "format": "npm run eslint -- --fix"
   },
   "engines": {
-    "node": ">=6"
+    "node": ">=10"
   },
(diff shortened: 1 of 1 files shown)


stianjensen:
I tried signing the cla, but something crashed 🤷

stianjensen:
> I tried signing the cla, but something crashed

Nvm, solved it by amending and force-pushing.

stianjensen:
I see that universalify is removed as a dependency altogether on the v5 branch, so this should not be needed then (since the updated node version requirements probably warrant a major release anyway).

PR #264, opened 2023-02-22
Typescript support

colincasey:
This PR adds two things to tough-cookie:
- TypeScript support
- A migration path from Vows to Jest for testing

The existing Vows test suite has been left in place to provide some level of assurance that these changes are (mostly) safe.  

So what's left to do here?
- [ ] tweaks to the TypeScript configuration
- [ ] changes or improvement to the defined types to be exported
- [ ] and any other misses that might come up in code review


lib/version.ts
```diff
@@ -0,0 +1,2 @@
+// TODO: this should be sourced from package.json (maybe?)
```
wjhsf, line 1: According to the [docs for `genversion`](https://www.npmjs.com/package/genversion), the whole purpose of this file is to _avoid_ bundling the `package.json`. Their main argument about the security implications isn't relevant to us, though, as this is an open source library, so the information is already available.

The only other consideration I can think of is whether it's worth including the whole large package.json just for the 5-character version.

lib/validators.ts
```diff
@@ -79,17 +80,11 @@ function validate(bool, cb, options) {
...
 }
 
-class ParameterError extends Error {
-  constructor(...params) {
+export class ParameterError extends Error {
+  constructor(...params: any[]) {
     super(...params);
+    if (Object.setPrototypeOf) {
+      Object.setPrototypeOf(this, ParameterError.prototype);
+    }
```
wjhsf, line 88: This shouldn't be necessary; the bug you're fixing here is an artifact of our tsconfig. We're currently targeting ES5, which predates classes. By changing to a newer target, the error disappears. (Changing to a newer target has other implications, though, so I think it's fine to leave for now. I have a follow up PR waiting where we can discuss this further.)

lib/version.ts
colincasey: and if you consider using `tough-cookie` in the context of a browser or native environment then we may not have access to the `package.json` file.  

the only place this is actually leveraged is during serialization of cookies.  i'm not sure that's entirely necessary but that's a discussion for another day.  i've reintroduced the version check test that originally led to this TODO being noted.

lib/validators.ts
colincasey: i did want to ask you about that.  if we're continually targeting modern versions of node and dropping support for versions outside of the supported LTS range we should tweak some of our settings to reflect that.

colincasey:
@awaterma @wjhsf i think this PR serves as a reasonable baseline to begin iterating on for v5

lib/validators.ts
wjhsf: I've done the work to target node 16 in #269.

wjhsf (approved):

PR #265, opened 2023-03-02
feat: removed psl package

YuryShkoda:


wjhsf:
Supporting [public suffixes](https://publicsuffix.org/learn/) is a critical security feature and is required by the [cookie spec](https://www.rfc-editor.org/rfc/rfc6265). The feature cannot be removed.

YuryShkoda:
> Supporting [public suffixes](https://publicsuffix.org/learn/) is a critical security feature and is required by the [cookie spec](https://www.rfc-editor.org/rfc/rfc6265). The feature cannot be removed.

Hi @wjhsf,
It is totally understandable that supporting public suffixes is a vital part of the project.
I should explain the intention, which is to reduce the bundle size of the project that is using `tough-cookie` package, but not using any features related to public suffixes. For example, [@sasjs/adapter](https://github.com/sasjs/adapter). `psl` package adds ~200Kb of hardcoded chars and it would be nice to be able to use a more lightweight version of `tough-cookie`.
I'm looking forward to hearing your thoughts.

colincasey:
Hi @YuryShkoda, I think for your use case you'll have to consider **forking** `tough-cookie` since we have no plans to break spec-compliance to achieve a smaller bundle size or **send us a PR** that swaps `psl` with an equivalent public suffix library with a lighter footprint.

PR #269, opened 2023-03-19
[v5] Update config

wjhsf:
I noticed that the class definition for `ParameterError` is more complex than necessary.
https://github.com/salesforce/tough-cookie/blob/069a7eebc3229ebb47db6650b022267b039f6145/lib/validators.ts#L83-L90
Updating that class definition uncovered a bug that was caused by `target: es5` in the tsconfig. Since one of our goals for v5 is to support node 16 as the minimum version, I updated the target to `ES2021`. This had some knock-on effects, like needing to update the `module` to support both our new jest tests and old vows tests. I also made a few other tweaks to improve our outputs while I was going through the config.


tsconfig.json
```diff
@@ -1,26 +1,27 @@
...
+    "noFallthroughCasesInSwitch": true,
     "noUncheckedIndexedAccess": true,
-    "importsNotUsedAsValues": "error",
+    "noImplicitOverride": true,
     "allowUnusedLabels": false,
-    "allowUnreachableCode": false,
-    "exactOptionalPropertyTypes": true
-  },
-  "include": ["./lib/**/*"],
-  "exclude": ["./lib/**/*.spec.ts"]
```
wjhsf, line 25: This is covered by the `rootDir` setting. Also, we should include test files when type checking, just exclude them when compiling/publishing.

tsconfig.json
```diff
@@ -1,26 +1,27 @@
...
-    "outDir": "./dist",
-    "allowJs": true,
-    "target": "es5",
-    "resolveJsonModule": true,
-    "esModuleInterop": true,
+    /* Language and Environment */
+    "target": "ES2021",
+    "lib": ["ES2021"],
+    /* Modules */
+    "module": "Node16",
```
wjhsf, line 7: Updated to meet our target node 16.

tsconfig.json
```diff
@@ -1,26 +1,27 @@
...
+    "rootDir": "./lib/",
+    /* Emit */
     "declaration": true,
+    "outDir": "./dist/",
```
wjhsf, line 11: Using `rootDir` ensures that we compile to `./dist`, rather than `./dist/lib`.

tsconfig.json
```diff
@@ -1,26 +1,27 @@
...
     "noUncheckedIndexedAccess": true,
-    "importsNotUsedAsValues": "error",
+    "noImplicitOverride": true,
     "allowUnusedLabels": false,
-    "allowUnreachableCode": false,
-    "exactOptionalPropertyTypes": true
-  },
-  "include": ["./lib/**/*"],
-  "exclude": ["./lib/**/*.spec.ts"]
+    "allowUnreachableCode": false
```
wjhsf, line 25: These are mostly just rearranged because I regenerated with `tsc --init`. Some keys have been removed that are covered by `strict: true`.

lib/validators.ts
```diff
@@ -27,6 +27,8 @@ SOFTWARE.
 ************************************************************************************ */
 "use strict";
 
+const toString = Object.prototype.toString
```
wjhsf, line 30: The `toString` used in this file (L54) is this method, accessed via the global object. That's weird, and TS complains.

lib/__tests__/ietf.spec.ts
```diff
@@ -31,9 +31,9 @@
 
 import {CookieJar, parseDate} from '../cookie'
 import url from 'url'
-import parserData from '../../test/ietf_data/parser.json'
-import bsdExampleDates from '../../test/ietf_data/dates/bsd-examples.json'
-import exampleDates from '../../test/ietf_data/dates/examples.json'
+import parserData from './data/parser'
+import bsdExampleDates from './data/dates/bsd-examples'
+import exampleDates from './data/dates/examples'
```
wjhsf, line 36: Changes to our tsconfig mean it's easier to have the test data within the `./lib` directory, and it's easier if they're TS files, rather than JSON.

jest.config.js
```diff
@@ -4,9 +4,6 @@ module.exports = {
   testEnvironment: "node",
   testRegex: "(/__tests__/.*|(\\.|/)(test|spec))\\.ts$",
   maxWorkers: 1,
-  globals: {
-    "ts-jest": {
-      isolatedModules: false
-    }
-  }
```
wjhsf, line 11: This is the default config, so we don't really need to specify it.

colincasey (approved):

PR #277, opened 2023-05-03
Fix error handling for SpecialUseDomains

dev-sharma-08:


wjhsf:
Because of the conditional on [line 51](https://github.com/salesforce/tough-cookie/blob/3e4d039b5e818e35d6b077e0bff52c32bd883d26/lib/pubsuffix-psl.js#L51), `allowSpecialUseDomain` should already only be `false` on the changed line, so this change does not seem necessary.

https://github.com/salesforce/tough-cookie/blob/3e4d039b5e818e35d6b077e0bff52c32bd883d26/lib/pubsuffix-psl.js#L51

dev-sharma-08:
@wjhsf, There is a case where the code flow can enter both the if block on line 51 and line 64 i.e. when
`domainPart.length = 1` and `topLevelDomain = SPECIAL_USE_DOMAINS - SPECIAL_TREATMENT_DOMAINS` i.e. `['local', 'example', 'test']`.
In this case it will throw error if we donot specify `!specialUseDomain` condition in second if block.

Issue #278, opened 2023-05-25
Can url-parse library be dropped?


DanEscudero:
Hello!

According to url-parse README file (https://www.npmjs.com/package/url-parse):

> url-parse was created in 2014 when the WHATWG URL API was not available in Node.js and the URL interface was supported only in some browsers. Today this is no longer true. The URL interface is available in all supported Node.js release lines and basically all browsers. Consider using it for better security and accuracy.

I wonder if url-parse is still necessary in this library, or it can be abandoned and URL be used instead. Is there any particular reason not to do this?
If this is ok, I'd be available to submit a Pull Request that does the work

awaterma:
Thank you @DanEscudero for pointing this out. We'd love a pull request if you'd like to submit a fix. We have all been really busy with our day jobs, so it may take a little while to review and merge. But we'd greatly appreciate the help! :)

wjhsf:
I believe that the `url-parse` library was originally added to support React Native, which does not have node's `url` library or a global `URL` object. I don't have any experience with React Native, so perhaps the situation has changed, but I assume that removing `url-parse` would cause things to break. We don't explicitly support React Native, but to avoid any unnecessary negative impacts we'd want to save this for a major release.

Issue #260, opened 2023-01-05
url-parse critical security patch


r4d:
url-parse security advisory update: https://github.com/advisories/GHSA-hgjh-723h-mx2j

Update url-parse > v1.5.8

Dremuchev:
Authorization Bypass Through User-Controlled Key in NPM url-parse prior to 1.5.9.

AndyDudleyAdvanced:
Hi,

Is there any reason why the reference to url-parse hasn't been updated to the patched version? (1.5.9)

colincasey:
@AndyDudleyAdvanced when you `npm install tough-cookie` it should resolve the `url-parse` dependency to version `1.5.10` since the range declared in [package.json](https://github.com/salesforce/tough-cookie/blob/master/package.json#L108) is `^1.5.3`.

colincasey:
Closing this issue since the version of `url-parse` resolved when installing `tough-cookie` is not affected by https://github.com/advisories/GHSA-hgjh-723h-mx2j (see the comment above).

PR #279, opened 2023-05-26
Remove url parse library

DanEscudero:
This PR aims to fix this issue:
https://github.com/salesforce/tough-cookie/issues/278

It also fixes this issue:
https://github.com/salesforce/tough-cookie/issues/260

This PR removes url-parse library and replaces url-parse library with native URL module.

Two commits were necessary to fix a couple issues found during tests:
- First test issue is related to IPv6 parsing. The original URL `http://[::ffff:127.0.0.1]/` parses to `http://[::ffff:7f00:1]/` according to IPv6 parsing specification (https://url.spec.whatwg.org/#concept-ipv6-parser); url-parse library parses this URL to `http://[::ffff:127.0.0.1]/`. In order to fix it, I changed the used domain in the test suite;

- Another test issue found is related to an invalid URL (`kyoto.jp`, without protocol). This issue is located in `CookieJar setCookie errors` batch. `url-parse` library allows it, but URL module does not, so I added a protocol in the domain.

### DIFF
--- a/lib/cookie.js
+++ b/lib/cookie.js
@@ -30,7 +30,6 @@
  */
 "use strict";
 const punycode = require("punycode/");
-const urlParse = require("url-parse");
 const pubsuffix = require("./pubsuffix-psl");
 const Store = require("./store").Store;
 const MemoryCookieStore = require("./memstore").MemoryCookieStore;
@@ -788,7 +787,7 @@ function getCookieContext(url) {
     // Silently swallow error
   }
 
-  return urlParse(url);
+  return new URL(url);
 }
 
 const cookieDefaults = {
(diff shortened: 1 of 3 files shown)


wjhsf:
The changes to the tests indicate that the `URL` is not a perfect drop-in replacement for `url-parse`. We don't want to introduce any breaking changes (also as discussed in https://github.com/salesforce/tough-cookie/issues/278#issuecomment-1597563622) in a minor release, so this change should be made to [v5](https://github.com/salesforce/tough-cookie/tree/v5) instead. Sorry for the mixup, @DanEscudero!

jimmywarting:
when could we expect a new major release (without url-parse)?
v5 branch seems to have lagged behind with 17 commits.

PR #275, opened 2023-04-18
Clean up validate function.

wjhsf:
To be merged after #271. Just a rewrite of the validate function to improve the signature and clean up the logic.

PR #288, opened 2023-06-23
Remove some @ts-ignore directives.

wjhsf:
Just plucking away at some of the @ts-ignore directives. I didn't address all of them in this PR; some I fixed in #275 and others I wasn't sure what the best solution would be.

### DIFF
--- a/lib/__tests__/cookieJar.spec.ts
+++ b/lib/__tests__/cookieJar.spec.ts
@@ -160,8 +160,9 @@ describe('CookieJar', () => {
     })
 
     it('should set a timestamp when storing or retrieving a cookie', async () => {
-      // @ts-ignore
-      cookie = Cookie.parse('a=b; Domain=example.com; Path=/')
+      // We know that we're passing a valid cookie, so we can use the non-null assertion
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      cookie = Cookie.parse('a=b; Domain=example.com; Path=/')!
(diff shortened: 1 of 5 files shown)


lib/__tests__/cookieJar.spec.ts
```diff
@@ -1145,11 +1145,8 @@ describe('loose mode', () => {
 it('should fix issue #132', async () => {
   const cookieJar = new CookieJar()
   await expect(
-    cookieJar.setCookie(
-      // @ts-ignore
-      { key: 'x', value: 'y' },
-      'http://example.com/',
-    ),
+    // @ts-expect-error test case is explicitly testing invalid input
```
wjhsf, line 1148: We _could_ do an `as any` assertion, but I think it makes more sense to use `@ts-expect-error` here because we are explicitly trying to do something invalid.

lib/cookie.ts
```diff
@@ -383,8 +383,8 @@ function canonicalDomain(str: string | null) {
 
 // S5.1.3 Domain Matching
 function domainMatch(
-  str?: string,
-  domStr?: string,
+  str?: string | null,
+  domStr?: string | null,
```
wjhsf, line 387: Based on the test cases, these methods seem to consider `null` to be valid input. Whether or not they _should_ going forward can be part of our upcoming review of how the library uses `null`/`undefined`.

colincasey (approved):

PR #289, opened 2023-07-06
Changes to support full eslint rule configurations

colincasey:
These are the remaining changes required to enable `plugin:@typescript-eslint/recommended-requiring-type-checking` as well as removing our other temporary rule overrides (`@typescript-eslint/ban-types`, `@typescript-eslint/no-explicit-any`, `@typescript-eslint/ban-ts-comment`, and `@typescript-eslint/no-unused-vars`)


lib/memstore.ts
```diff
@@ -188,12 +205,23 @@ export class MemoryCookieStore extends Store {
       return promiseCallback.promise
     }
 
-    const domainEntry: { [key: string]: any } =
-      this.idx[domain] ?? Object.create(null)
+    const domainEntry: {
+      [path: string]: {
+        [key: string]: Cookie
+      }
+    } =
```
wjhsf, line 212:
```suggestion
    const domainEntry: MemoryCookieStoreIndex[string] =
```

lib/memstore.ts
```diff
@@ -188,12 +205,23 @@ export class MemoryCookieStore extends Store {
...
+      this.idx[domain] ??
+      (Object.create(null) as {
+        [path: string]: {
+          [key: string]: Cookie
+        }
+      })
```
wjhsf, line 218:
```suggestion
      Object.create(null)
```
Pretty sure the type assertion shouldn't be necessary. But if it is, use `MemoryCookieStoreIndex[string]` instead.

wjhsf (approved):

lib/pubsuffix-psl.ts
```diff
@@ -58,11 +58,12 @@ export function getPublicSuffix(
 
   if (
     allowSpecialUseDomain &&
-    typeof topLevelDomain === 'string' &&
+    topLevelDomain !== undefined &&
```
wjhsf, line 61: Why this switch?

lib/memstore.ts
```diff
@@ -39,24 +39,30 @@ import {
...
+
+export class MemoryCookieStore extends Store {
+  override synchronous: boolean
+  idx: MemoryCookieStoreIndex
 
   constructor() {
     super()
     this.synchronous = true
-    this.idx = Object.create(null)
+    this.idx = Object.create(null) as MemoryCookieStoreIndex
```
wjhsf, line 57:
```suggestion
    // `Object.create` returns `any`, but we know it's safe to use here
    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
    this.idx = Object.create(null)
```
[nit] Since it's eslint that's complaining, I'd fix it at the eslint level, rather than using a type assertion.

lib/memstore.ts
```diff
@@ -39,24 +39,30 @@ import {
...
     const customInspectSymbol = getCustomInspectSymbol()
     if (customInspectSymbol) {
-      // @ts-ignore
-      this[customInspectSymbol] = this.inspect
+      Object.defineProperty(this, customInspectSymbol, {
+        value: this.inspect.bind(this),
```
wjhsf, line 61: Why use a bound function rather than the original one?

lib/cookie.ts
```diff
@@ -1235,10 +1298,7 @@ const defaultGetCookieOptions: GetCookiesOptions = {
 }
 
 export function createPromiseCallback<T>(args: IArguments): PromiseCallback<T> {
```
wjhsf, line 1300: The only thing we use `args` for is to extract the callback from the last one. Since nobody likes `arguments` any more, maybe we should just directly pass the callback.

lib/cookie.ts
```diff
@@ -1247,64 +1307,78 @@ export function createPromiseCallback<T>(args: IArguments): PromiseCallback<T> {
...
-  if (typeof args[args.length - 1] === 'function') {
-    const cb = args[args.length - 1]
+  const cb = args[args.length - 1] as unknown
+  if (typeof cb === 'function') {
     callback = (err, result) => {
       try {
         cb(err, result)
       } catch (e) {
-        reject(e instanceof Error ? e : new Error(`${e}`))
+        reject(e instanceof Error ? e : new Error())
```
wjhsf, line 1316: Why this change? We should probably have some kind of message to be a bit more helpful, and maybe include the original error. (Also returning so we safely short-circuit.)
```suggestion
        if (e instanceof Error) return reject(e)
        if (typeof e !== 'object' || e === null) return reject new Error(e)
        const wrappedError = new Error('An unknown error occurred.')
        wrappedError.original = e
        return reject(wrappedError)
```

wjhsf (commented):
👍, all my comments are just nits.

lib/memstore.ts
colincasey: it was the only way to eliminate the `@ts-ignore`

lib/cookie.ts
colincasey: i agree in principal but, the way the code is organized with optional callbacks at various positions in the arguments list for many different functions, this approach seemed less disruptive.  

and going forward i'd like for us to drop callbacks entirely and eliminate all the extra complexity that comes with managing this promise/callback hybrid style in the codebase.

lib/pubsuffix-psl.ts
colincasey: it was either a linter or IDE suggestion

wjhsf (approved):

PR #297, opened 2023-08-27
fix #256 

### LINKED ISSUE #256: Trailing slash breaks ESM compatibility
Originally reported in https://github.com/sveltejs/kit/issues/7085

Loading this library in a project that uses Node.js ESM without bundling does not work:

> Error [ERR_UNSUPPORTED_DIR_IMPORT]: Directory import 'node_modules/punycode/' is not supported resolving ES modules imported from build/server/chunks/_server.ts-7b9e82e7.js

I believe the error is coming from this line:

https://github.com/salesforce/tough-cookie/blob/d4ac5801dd2c2d53eec51329e5380bbffb23bfaf/lib/cookie.js#L32

Is there a reason to have that trailing slash or can it be removed?

hrueger:
by importing `.js` file and therefore removing the trailing slash

I'm also using Svelte Kit and currently patching this via `yarn patch`.


hrueger:
I signed the CLA, but then I got an error. Can you please recheck?

hrueger:
Hm, did I break the tests? That's strange...

awaterma:
@hrueger -- can you try signing the CLA again? :) Hope you are doing well!

hrueger:
Yes, thanks. When clicking on the link and logging in with GitHub, I get this:
grafik

What to do?

awaterma (commented):
This change looks good to me; but there's a file conflict. Can you please update your pull request @hrueger? Thank you!

hrueger:
@awaterma Now that the project is written in typescript, I get ts errors when importing from that javascript file since the `@types/punycode` packages seems to only support the index export... any idea?

colincasey:
@hrueger it seems like the userland `punycode` module doesn't play well with Typescript (https://github.com/mathiasbynens/punycode.js/issues/132). 

My suggested workaround would be to:

- remove `@types/punycode` from `package.json` 
- add a custom declaration file like `lib/punycode.d.ts` which should look something like this:

    ```ts
    declare module 'punycode/punycode.js' {
      // copy the types declared here:
      // https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/punycode/index.d.ts
    }
    ```

- change the `punycode/` import to `punycode/punycode.js`

And, since we only use a very small part of the API from `punycode`, you could possibly just include the type declaration for `toASCII` and change the import to `import { toASCII } from 'punycode/punycode.js'`.

hrueger:
@colincasey I submitted a PR to DefinitelyTyped: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/68273
Let's see if this makes it through ;-)

hrueger:
@colincasey My DefinitelyTyped PR was merged earlier, so I updated the package and now the types should be fine 🎉

colincasey (approved):

PR #302, opened 2023-09-14
Update url-parse to latest version to mitigate Security issue.

AndyDudleyAdvanced:
Fixes https://github.com/advisories/GHSA-hgjh-723h-mx2j

referenced in this issue:

https://github.com/salesforce/tough-cookie/issues/260

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,6 +105,6 @@
     "psl": "^1.1.33",
     "punycode": "^2.1.1",
     "universalify": "^0.2.0",
-    "url-parse": "^1.5.3"
+    "url-parse": "^1.5.10"
   }
 }


awaterma:
@colincasey -- should we consider this p/r or a major version update for 5.0?

colincasey:
@awaterma I've already commented on #260 as to why this isn't a security issue for our published version of `tough-cookie` but I will restate it here.  When you `npm install tough-cookie` it currently resolves the `url-parse` dependency to version `1.5.10` which is not affected by the vulnerability noted in this PR since the range declared in [package.json](https://github.com/salesforce/tough-cookie/blob/master/package.json#L108) is `^1.5.3`.

PR #316, opened 2023-10-27
Avoid using `arguments`

wjhsf:
Using `arguments` is a known code smell. It is used by `createPromiseCallback` to extract the callback from the last provided argument. However, all calling code already knows which variable contains the callback, so we don't actually need to use `arguments`.

### DIFF
--- a/lib/__tests__/removeAll.spec.ts
+++ b/lib/__tests__/removeAll.spec.ts
@@ -154,7 +154,7 @@ class StoreWithoutRemoveAll extends Store {
     if (!callback) {
       throw new Error('This should not be undefined')
     }
-    return callback(undefined, undefined)
+    return callback(null, undefined)
   }
 
   override findCookies(
@@ -177,7 +177,7 @@ class StoreWithoutRemoveAll extends Store {
     if (!callback) {
       throw new Error('This should not be undefined')
     }
-    return callback(undefined, [])
+    return callback(null, [])
   }
 
   override putCookie(cookie: Cookie): Promise<void>
@@ -188,7 +188,7 @@ class StoreWithoutRemoveAll extends Store {
     if (!callback) {
       throw new Error('This should not be undefined')
     }
-    return callback(undefined)
+    return callback(null)
   }
 
   override getAllCookies(): Promise<Cookie[]>
@@ -198,7 +198,7 @@ class StoreWithoutRemoveAll extends Store {
     if (!callback) {
       throw new Error('This should not be undefined')
     }
-    return callback(undefined, this.cookies.slice())
+    return callback(null, this.cookies.slice())
   }
 
   override removeCookie(
@@ -222,7 +222,7 @@ class StoreWithoutRemoveAll extends Store {
     if (!callback) {
       throw new Error('This should not be undefined')
     }
-    return callback(undefined)
+    return callback(null)
   }
 }
 
--- a/lib/__tests__/removeAll.spec.ts
+++ b/lib/__tests__/removeAll.spec.ts
@@ -138,18 +138,18 @@ class StoreWithoutRemoveAll extends Store {
     domain: string,
     path: string,
     key: string,
-  ): Promise<Cookie>
+  ): Promise<Cookie | undefined>
   override findCookie(
     domain: string,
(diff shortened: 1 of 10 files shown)


lib/memstore.ts
```diff
@@ -72,43 +72,25 @@ export class MemoryCookieStore extends Store {
     domain: string | null,
     path: string | null,
     key: string | undefined,
-  ): Promise<Cookie | null | undefined>
+  ): Promise<Cookie | undefined>
```
wjhsf, line 75: There was only one path using `null`, so I changed it to `undefined` for consistency.

lib/memstore.ts
```diff
@@ -72,43 +72,25 @@ export class MemoryCookieStore extends Store {
...
       return promiseCallback.resolve(undefined)
     }
-
-    if (key == null) {
-      return promiseCallback.resolve(null)
-    }
-
-    cb(null, pathEntry[key] || null)
-    return promiseCallback.promise
+    const result = this.idx?.[domain]?.[path]?.[key]
```
wjhsf, line 92: Modern syntax makes this function much smaller!

lib/__tests__/removeAll.spec.ts
```diff
@@ -138,23 +138,23 @@ class StoreWithoutRemoveAll extends Store {
...
     key: string,
-    callback: Callback<Cookie>,
+    callback: Callback<Cookie | undefined>,
   ): void
   override findCookie(
     _domain: string,
     _path: string,
     _key: string,
-    callback?: Callback<Cookie>,
+    callback?: Callback<Cookie | undefined>,
```
wjhsf, line 152: With changes to the callback signature, it's the responsibility of calling code to permit `undefined`.

lib/__tests__/removeAll.spec.ts
```diff
@@ -138,23 +138,23 @@ class StoreWithoutRemoveAll extends Store {
...
   ): unknown {
     if (!callback) {
       throw new Error('This should not be undefined')
     }
-    return callback(undefined, undefined)
+    return callback(null, undefined)
```
wjhsf, line 157: Callback signature has been standardized to only use `null` for "no error".

lib/cookie/cookieJar.ts
```diff
@@ -186,36 +188,42 @@ export class CookieJar {
...
-    options?: SetCookieOptions | Callback<Cookie>,
-    callback?: Callback<Cookie>,
+    options?: SetCookieOptions | Callback<Cookie | undefined>,
+    callback?: Callback<Cookie | undefined>,
   ): unknown {
-    const promiseCallback = createPromiseCallback<Cookie>(arguments)
+    if (typeof options === 'function') {
+      callback = options
+      options = undefined
+    }
```
wjhsf, line 225: Since we have mushed up function signatures, we have to rearrange the parameters to use them correctly.

lib/cookie/cookieJar.ts
```diff
@@ -261,7 +269,9 @@ export class CookieJar {
       const parsedCookie = Cookie.parse(cookie.toString(), { loose: loose })
       if (!parsedCookie) {
         err = new Error('Cookie failed to parse')
-        return promiseCallback.reject(options?.ignoreError ? null : err)
+        return options?.ignoreError
+          ? promiseCallback.resolve(undefined)
+          : promiseCallback.reject(err)
```
wjhsf, line 274: There are a lot of places where we have `ignoreError` resulting in a rejection, but that's not ignoring the error! That's just changing the root cause into a mystery.

lib/cookie/cookieJar.ts
```diff
@@ -416,17 +435,19 @@ export class CookieJar {
...
         return
       }
 
-      const next = function (err: Error | undefined): void {
-        if (err || typeof cookie === 'string') {
+      const next = function (err: Error | null): void {
+        if (err) {
           cb(err)
+        } else if (typeof cookie === 'string') {
+          cb(null, undefined)
```
wjhsf, line 450: A quirk of the updated callback signature.

awaterma (approved):
Approved.

PR #323, opened 2023-11-03
Bump dependencies.

wjhsf:
Bumps everything to latest except @types/node (we want to support old versions) and TypeScript (it's a special case because we're a library). Changed a few tests to work with latest jest, and updated the lint config based on what's available in the latest versions (plus associated fixes).


lib/__tests__/parse.spec.ts
```diff
@@ -455,22 +461,3 @@ describe('Cookie.parse', () => {
     },
   )
 })
-
-// way too many spaces - takes about the same time for each
-it('should parse a long cookie string with spaces in roughly the same amount of time as one with short spaces', () => {
```
wjhsf, line 460: This was a duplicate of the prior block of tests.

lib/__tests__/parse.spec.ts
```diff
@@ -307,7 +307,7 @@ describe('Cookie.parse', () => {
     },
     // way too many semicolons followed by non-semicolon
     {
-      input: `foo=bar${';'.repeat(65535)} domain=example.com`,
+      input: `foo=bar<REPEAT ;> domain=example.com`,
```
wjhsf, line 310: Printing 65535 characters in the jest output is not that fun to deal with.

colincasey (approved):

PR #345, opened 2023-12-13
Fix expiry time not updating when cookie is updated

### LINKED ISSUE #154: CookieJar.setCookie does not slide expiration for set-cookie headers with max-age
The creation time from old-cookie is carried forward per the spec, but this causes the expiryTime() to be unaffected by subsequent set-cookie headers, so the cookie will expire when its creation + max-age > now, even if subsequent set-cookie headers specified max-age with the intent of making the expiration now + max-age.

I'm currently working around this by just having a Store implementation that passes `null` to the callback, so that the creation is not carried forward.

### LINKED ISSUE #343: CookieJar.setCookie does not slide expiration for set-cookie headers with max-age
> My reading of RFC6265 is that `expiry-time` and `creation-time` are two distinct fields (Section 5.3), and `creation-time` is not used for expiration at all. Steps 5.3.2 and 5.3.11.3 cover `creation-time` and `expiry-time` is covered by 5.3.3, and sections 5.2.1 and 5.2.2.

As far as I can tell, this issue still exists, assuming my past interpretation of RFC6265 was correct.

_Originally posted by @VelynM in https://github.com/salesforce/tough-cookie/issues/154#issuecomment-1839869846_

colincasey:
The `cookie.expiryTime()` method was calculating an `expiry-time` value using `creation-time + max-age` if the `max-age` value was set. This is not correct as RFC6265 does not say anything about `creation-time` having any effect on `max-age`.

The main issue is that `max-age` should be stored as a date (see text below), not an offset number as it currently is implemented.

> 5.2.2.  The Max-Age Attribute
>
> If the attribute-name case-insensitively matches the string "Max-
> Age", the user agent MUST process the cookie-av as follows.
>
> If the first character of the attribute-value is not a DIGIT or a "-"
> character, ignore the cookie-av.
>
> If the remainder of attribute-value contains a non-DIGIT character,
> ignore the cookie-av.
>
> Let delta-seconds be the attribute-value converted to an integer.
>
> If delta-seconds is less than or equal to zero (0), let expiry-time
> be the earliest representable date and time.  Otherwise, **let the
> expiry-time be the current date and time plus delta-seconds seconds**.
>
> Append an attribute to the cookie-attribute-list with an attribute-
> name of Max-Age and an attribute-value of expiry-time.

Because `max-age` is an offset, the `expiry-time` now needs to be calculated from "some" point in time and `creation-time` was chosen. In many cases this works fine but, if you have an existing cookie stored with a `max-age` attribute and then update it, the `expiry-time` will be incorrectly report the previous value's `expiry-time`. This is due to the following rule regarding `creation-time` from the spec:

> 5.3.  Storage Model
>
> 11.  If the cookie store contains a cookie with the same name,
>      domain, and path as the newly created cookie:
>
>      1.  Let old-cookie be the existing cookie with the same name,
>          domain, and path as the newly created cookie.  (Notice that
>          this algorithm maintains the invariant that there is at most
>          one such cookie.)
>
>      2.  If the newly created cookie was received from a "non-HTTP"
>          API and the old-cookie's http-only-flag is set, abort these
>          steps and ignore the newly created cookie entirely.
>
>      3.  Update the creation-time of the newly created cookie to
>          match the creation-time of the old-cookie.
>
>      4.  Remove the old-cookie from the cookie store.

This should make it clear that `creation-time` should not be used to calculate the `expiry-time`. Given that we still need to calculate the `expiry-time` due to how `max-age` is stored, the better value to use in this circumstance is the `last-access-time` which is always updated when `setCookie(...)` is called.

Fixes #343
Fixes #154

### DIFF
--- a/lib/__tests__/cookieJar.spec.ts
+++ b/lib/__tests__/cookieJar.spec.ts
@@ -1203,6 +1203,48 @@ it('should fix issue #282 - Prototype pollution when setting a cookie with the d
   expect('/notauth' in pollutedObject).toBe(false)
 })
 
+it('should fix issue #154 - Expiry should not be affected by creation date', async () => {
+  // setting the time to zero here just to make it obvious that the expiry time is updated
+  jest.useFakeTimers({ now: 0 })
+
+  const jar = new CookieJar()
+
+  await jar.setCookie('foo=bar; Max-Age=60;', 'https://example.com')
+
+  const initialCookies = await jar.getCookies('https://example.com')
+  expect(initialCookies).toEqual([
+    expect.objectContaining({
+      key: 'foo',
+      value: 'bar',
+      path: '/',
+      domain: 'example.com',
+      maxAge: 60,
+    }),
+  ])
+  // the expiry time should be 60s from now (0)
+  expect(initialCookies[0]?.expiryTime()).toBe(60 * 1000)
+
+  // advance the time by 1s, so now = 1000
+  jest.advanceTimersByTime(1000)
+
+  await jar.setCookie('foo=bar; Max-Age=60;', 'https://example.com')
+
+  const updatedCookies = await jar.getCookies('https://example.com')
+  expect(updatedCookies).toEqual([
+    expect.objectContaining({
+      key: 'foo',
+      value: 'bar',
+      path: '/',
+      domain: 'example.com',
+      maxAge: 60,
+      // the creation time should be unchanged as per the spec
+      creation: initialCookies[0]?.creation,
+    }),
+  ])
+  // the expiry time should be 60s from now (1000)
+  expect(updatedCookies[0]?.expiryTime()).toBe(60 * 1000 + 1000)
+})
+
 // special use domains under a sub-domain
 describe.each(['local', 'example', 'invalid', 'localhost', 'test'])(
   'when special use domain is dev.%s',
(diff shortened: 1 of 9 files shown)


lib/__tests__/cookieJar.spec.ts
```diff
@@ -1203,6 +1203,48 @@ it('should fix issue #282 - Prototype pollution when setting a cookie with the d
   expect('/notauth' in pollutedObject).toBe(false)
 })
 
+it('should fix issue #154 - Expiry should not be affected by creation date', async () => {
+  // setting the time to zero here just to make it obvious that the expiry time is updated
+  jest.useFakeTimers({ now: 0 })
```
wjhsf, line 1208: Do we need to use real timers after the test, or is this scoped to the test?

wjhsf (approved):

lib/__tests__/cookieJar.spec.ts
colincasey: are you asking in case we need to restore the timers after the test run?  because the test suite relies on mocking out the timers but this is the only one that sets `now` to 0.

awaterma (approved):

lib/__tests__/cookieJar.spec.ts
colincasey: @wjhsf i've enabled the fake timers globally in the jest config since we have no tests that rely on a real timers in our jest suite.

wjhsf (approved):

PR #346, opened 2023-12-13
Replace `psl` with `tldts`

### LINKED ISSUE #239: expensive blocking PSL lookup

### LINKED ISSUE #338: Replace psl dependency with tldts
The [psl](https://github.com/lupomontero/psl) project seems to be [dead](https://github.com/lupomontero/psl/pulls).

It's generating a [punycode deprecation warning](#324) despite an 18mo old open pull request fixing this issue.  tough-cookie incurs a serious [performance bottleneck](#239) due to its use of psl.

tldts is [active](https://github.com/remusao/tldts/pulls?q=is%3Apr+is%3Amerged+), and [much, much more performant](https://remusao.github.io/posts/tldts-benchmarks.html). 

Can you fix this, please?

colincasey:
One of the topics that came up in discussions for v5 was considering replacements for `psl` (see https://github.com/salesforce/tough-cookie/discussions/255). This PR replaces that library with `tldts` which is more performant.

Fixes #239
Fixes #338

### DIFF
--- a/lib/pubsuffix-psl.ts
+++ b/lib/pubsuffix-psl.ts
@@ -29,7 +29,7 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 'use strict'
-import * as psl from 'psl'
+import { getDomain } from 'tldts'
 
 // RFC 6761
 const SPECIAL_USE_DOMAINS = ['local', 'example', 'invalid', 'localhost', 'test']
@@ -84,5 +84,5 @@ export function getPublicSuffix(
     )
   }
 
-  return psl.get(domain)
+  return getDomain(domain)
 }
--- a/lib/cookie/cookie.ts
+++ b/lib/cookie/cookie.ts
@@ -32,7 +32,7 @@
 // This file was too big before we added max-lines, and it's ongoing work to reduce its size.
 /* eslint max-lines: [1, 750] */
 
-import * as pubsuffix from '../pubsuffix-psl'
+import { getPublicSuffix } from '../getPublicSuffix'
 import * as validators from '../validators'
 import { getCustomInspectSymbol } from '../utilHelper'
 import { inOperator } from '../utils'
@@ -559,7 +559,7 @@ export class Cookie {
       if (cdomain.match(/\.$/)) {
         return false // S4.1.2.3 suggests that this is bad. domainMatch() tests confirm this
       }
-      const suffix = pubsuffix.getPublicSuffix(cdomain)
+      const suffix = getPublicSuffix(cdomain)
       if (suffix == null) {
         // it's a public suffix
         return false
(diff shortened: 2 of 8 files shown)


lib/pubsuffix-psl.ts
```diff
@@ -84,5 +84,5 @@ export function getPublicSuffix(
     )
   }
 
-  return psl.get(domain)
+  return getDomain(domain)
```
remusao, line 87: Note: by default `tldts` ignores the private section of the public suffix list. If the intent here is to also take into account these rules (which I assume might be the case since the `tough-cookie` library deals with cookies and browsers take into account the private section of the public suffix list) then you'd need to add the following option while calling `getDomain`:
```suggestion
  return getDomain(domain, { allowPrivateDomains: true })
```

Additionally, `tldts` is able to handle arbitrary inputs (i.e. both URLs and hostnames alike) with a slight overhead for the detection. If the input is already a valid hostname (e.g. if it was extracted using a `URL` parser) you can squeeze a bit of extra performance using the `{ extractHostname: false }` option, which skips the logic of extracting hostnames  from URL inputs as well as hostname validation.

Lastly, `tldts` is also able to handle IPv4 and IPv6 inputs by default, which I believe `psl` does not support. So if these inputs are already handled in a different way you can also disable this verification in `tldts` using the option: `{ detectIp: false }` (which has some small overhead).

wjhsf: Thanks for all those tips!

awaterma, line 87:
```suggestion
  return getDomain(domain, { allowPrivateDomains: true })
```

I think we should follow through with @remusao's suggestion. :)

awaterma (approved):
LGTM.

ravanscafi:
Hey folks, could you please tag a new version of the package? We are still getting PSL errors because the last version is from june 2023. 

thanks!

colincasey:
@ravanscafi These latest features are for our an upcoming `v5` release.  Pre-releases are currently being tagged with `@next` in npm which can be installed with:

```sh
npm install tough-cookie@next
```

We will be publishing a new pre-release with this `tldts` change included by the end of this week.  

Do you plan on trying out the pre-release or will you be waiting until an official `v5` version is ready?

PR #349, opened 2023-12-23
docs: use correct memstore file link

alissonsleal:
Now `lib/memstore.ts` is linked correctly in the documentation after it was migrated to typescript


wjhsf (approved):
Thanks for catching this!

PR #359, opened 2024-01-17
Remove workarounds for `util` in non-node environments

wjhsf:
Rather than having conditional logic, this simplifies the code by always using the custom inspect symbol and always using our custom inspect function for `MemoryCookieStore`.


lib/cookie/cookie.ts
```diff
@@ -421,16 +420,12 @@ export class Cookie {
...
-    const customInspectSymbol = getCustomInspectSymbol()
-    if (customInspectSymbol) {
-      Object.defineProperty(this, customInspectSymbol, {
-        value: this.inspect.bind(this),
-        enumerable: false,
-        writable: false,
-        configurable: false,
-      })
-    }
+    Object.defineProperty(this, NODEJS_UTIL_INSPECT_CUSTOM, {
```
wjhsf, line 423: One change that we could do, but I wanted to discuss before deciding, is whether to move this to be a regular prop on the prototype, rather than a weird bound prop on the instance. Technically a breaking change, but a small one, and we're gonna land a major anyway, so we could get away with it.

lib/memstore.ts
```diff
@@ -57,20 +57,16 @@ export class MemoryCookieStore extends Store {
...
-    const customInspectSymbol = getCustomInspectSymbol()
-    if (customInspectSymbol) {
-      Object.defineProperty(this, customInspectSymbol, {
-        value: this.inspect.bind(this),
-        enumerable: false,
-        writable: false,
-        configurable: false,
-      })
-    }
+    Object.defineProperty(this, NODEJS_UTIL_INSPECT_CUSTOM, {
```
wjhsf, line 60: One change that we could do, but I wanted to discuss before deciding, is whether to move this to be a regular prop on the prototype, rather than a weird bound prop on the instance. Technically a breaking change, but a small one, and we're gonna land a major anyway, so we could get away with it.

colincasey:
@wjhsf another option to consider if we're going to make a breaking change is to remove the `inspect` functions entirely.

colincasey (approved):

awaterma (approved):
Approved!

PR #366, opened 2024-02-20
Handle unlikely edge case in unimportant util

wjhsf:
Currently `safeToString([Object.create(null)])` throws, even though it shouldn't. This PR fixes that.

### DIFF
--- a/lib/__tests__/util.spec.ts
+++ b/lib/__tests__/util.spec.ts
@@ -1,0 +1,28 @@
+import { safeToString } from '../utils'
+
+describe('safeToString', () => {
+  const basic = [
+    undefined,
+    null,
+    true,
+    'string',
+    123,
+    321n,
+    { object: 'yes' },
+    [1, 'hello', true, null],
+    (a: number, b: number) => a + b,
+    Symbol('safeToString'),
+  ]
+  const testCases = [
+    ...basic.map((input) => [input, String(input)]),
+    [Object.create(null), '[object Object]'],
+    [
+      [Object.create(null), Symbol('safeToString')],
+      '[object Object],Symbol("safeToString")',
+    ],
+  ]
+
+  it.each(testCases)('works on %s', (input, output) => {
+    expect(safeToString(input)).toBe(String(output))
+  })
+})
--- a/lib/__tests__/util.spec.ts
+++ b/lib/__tests__/util.spec.ts
@@ -18,7 +18,7 @@ describe('safeToString', () => {
     [Object.create(null), '[object Object]'],
     [
       [Object.create(null), Symbol('safeToString')],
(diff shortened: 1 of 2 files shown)


lib/utils.ts
```diff
@@ -14,12 +14,16 @@ export const objectToString = (obj: unknown) =>
...
 /** Safely converts any value to string, using the value's own `toString` when available. */
-export const safeToString = (val: unknown) => {
-  // Ideally, we'd just use String() for everything, but it breaks if `toString` is missing (mostly
-  // values with no prototype), so we have to use Object#toString as a fallback.
+export const safeToString = (val: unknown): string => {
+  // Using .toString() fails for null/undefined and implicit conversion (val + "") fails for symbols
+  // and objects with null prototype
   if (val === undefined || val === null || typeof val.toString === 'function') {
     return String(val)
+  } else if (Array.isArray(val)) {
```
colincasey, line 22: In the `Array` case, this 2nd conditional will never be reached because of the 1st conditional which includes `typeof val.toString === 'function'` which is true for `Array` instances.  I would suggest swapping the order here so the `isArray` check happens first.

lib/__tests__/util.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+    'string',
+    123,
+    321n,
+    { object: 'yes' },
+    [1, 'hello', true, null],
+    (a: number, b: number) => a + b,
+    Symbol('safeToString'),
+  ]
+  const testCases = [
+    ...basic.map((input) => [input, String(input)]),
```
colincasey, line 17: Instead of mapping the `basic` values and relying on `String(input)` to build the `[input, output]` pair, maybe this should explicitly declare the values.  E.g.;

```js
const testCases = [
  [undefined, 'undefined'],
  [null, 'null'],
  [true, 'true'],
  ['string', 'string'],
  ...
  [
    [Object.create(null), Symbol('safeToString')],
    '[[object Object], Symbol(safeToString)]',
  ],
]
```

lib/utils.ts
```diff
@@ -14,12 +14,16 @@ export const objectToString = (obj: unknown) =>
...
+    // Array#toString implicitly converts its values to strings, which is what we're trying to avoid
+    return val.map(safeToString).join()
```
colincasey, line 24: The output of the `Array` case should wrap the inner values with `[]`.  

```suggestion
    return `[${val.map(safeToString).join(', ')}]`
```

colincasey (changes requested):

lib/utils.ts
wjhsf: This is for `toString`, not JSON, so we don't want brackets or spaces.

```sh
$ node -p '[1, true, "foo"].join()'
1,true,foo
```

PR #377, opened 2024-03-01
clean up usage of null/undefined

wjhsf:
This standardizes most of our methods to accept both null and undefined, and mostly return a value or undefined. Exceptions that I noticed are `canonicalDomain`, `getPublicSuffix`, `permuteDomain`, and `domainMatch`, which all return null. I don't remember why I skipped them when I originally did this, so I figured I'd just leave them for now. (I did most of this work a while ago, and evidently never pushed!)

This also introduces a helper type `Nullable<T>` for readability.


lib/cookie/cookie.ts
```diff
@@ -382,46 +403,42 @@ const cookieDefaults = {
...
-  key: string | undefined
-  value: string | undefined
-  expires: Date | 'Infinity' | null | undefined
-  maxAge: number | 'Infinity' | '-Infinity' | undefined
-  domain: string | null | undefined
-  path: string | null | undefined
-  secure: boolean | undefined
-  httpOnly: boolean | undefined
-  extensions: string[] | null | undefined
+  key: string
```
wjhsf, line 409: All of these props seem to have `| undefined`, but not actually need it, because none of the `cookieDefaults` are `undefined`.

lib/cookie/cookie.ts
```diff
@@ -382,46 +403,42 @@ const cookieDefaults = {
...
   creation: Date | 'Infinity' | null
-  creationIndex: number | undefined
-  hostOnly: boolean | null | undefined
-  pathIsDefault: boolean | null | undefined
-  lastAccessed: Date | 'Infinity' | null | undefined
+  creationIndex: number
+  hostOnly: boolean | null
+  pathIsDefault: boolean | null
+  lastAccessed: Date | 'Infinity' | null
   sameSite: string | undefined
```
wjhsf, line 423: `sameSite` _does_ actually use `undefined` to mean "no value". Which is anomalous, but more of a change than I'd want to deal with, here.

lib/cookie/cookie.ts
```diff
@@ -382,46 +403,42 @@ const cookieDefaults = {
...
+    this.domain = options.domain ?? cookieDefaults.domain
+    this.path = options.path ?? cookieDefaults.path
+    this.secure = options.secure ?? cookieDefaults.secure
+    this.httpOnly = options.httpOnly ?? cookieDefaults.httpOnly
+    this.extensions = options.extensions ?? cookieDefaults.extensions
+    this.creation = options.creation ?? cookieDefaults.creation
+    this.hostOnly = options.hostOnly ?? cookieDefaults.hostOnly
+    this.pathIsDefault = options.pathIsDefault ?? cookieDefaults.pathIsDefault
+    this.lastAccessed = options.lastAccessed ?? cookieDefaults.lastAccessed
+    this.sameSite = options.sameSite ?? cookieDefaults.sameSite
```
wjhsf, line 439: TypeScript complains if we use `Object.assign`, so I just made it all explicit.

colincasey:
@wjhsf I'm finding this PR a bit too hard to review. Now that we've done most of the major cleanup with the migration to TypeScript I think this would be easier to review as separate, smaller PRs for each of the included changes such as:
* Clean up cookie creation options
* Prefer promiseCallback.resolve where possible
* Introduction and usage of the `Nullable` type
* etc.

wjhsf:
Replaced by #380 #381.

PR #380, opened 2024-03-13
clean up usage of null/undefined

wjhsf:
This standardizes most of our methods to accept both null and undefined, and mostly return a value or undefined. Exceptions that I noticed are `canonicalDomain`, `getPublicSuffix`, `permuteDomain`, and `domainMatch`, which all return null. I don't remember why I skipped them when I originally did this, so I figured I'd just leave them for now. (I did most of this work a while ago, and evidently never pushed!)

This also introduces a helper type `Nullable<T>` for readability.

nb: #377 but without changes to cookie creation

### DIFF
--- a/lib/cookie/cookie.ts
+++ b/lib/cookie/cookie.ts
@@ -30,7 +30,7 @@
  */
 
 // This file was too big before we added max-lines, and it's ongoing work to reduce its size.
-/* eslint max-lines: [1, 750] */
+/* eslint max-lines: [1, 800] */
 
 import { getPublicSuffix } from '../getPublicSuffix'
 import * as validators from '../validators'
--- a/lib/cookie/cookie.ts
+++ b/lib/cookie/cookie.ts
@@ -280,6 +280,10 @@ function parse(str: string, options?: ParseCookieOptions): Cookie | undefined {
   return c
 }
 
+// TBD: `null` is valid JSON, but `undefined` is not. Options:
+// 1. Change this to return `undefined` - weird because it's not JSON
+// 2. Keep this as `null` - weird because it's a violation of our new convention
+// 3. Change *everything* to return `null` - a lot of work, maybe other edge cases that we can't change?
 function fromJSON(str: unknown): Cookie | null {
   if (!str || validators.isEmptyString(str)) {
     return null
@@ -532,6 +536,7 @@ export class Cookie {
     return obj
   }
 
+  // TBD: This is a wrapper for `fromJSON`, return type depends on decision there
   clone(): Cookie | null {
     return fromJSON(this.toJSON())
   }
@@ -697,15 +702,12 @@ export class Cookie {
   }
 
   // Mostly S5.1.2 and S5.2.3:
-  canonicalizedDomain(): string | null {
-    if (this.domain == null) {
-      return null
-    }
+  canonicalizedDomain(): string | undefined {
     return canonicalDomain(this.domain)
   }
 
-  cdomain(): string | null {
-    return this.canonicalizedDomain()
+  cdomain(): string | undefined {
+    return canonicalDomain(this.domain)
   }
 
   static parse = parse
--- a/lib/cookie/cookie.ts
+++ b/lib/cookie/cookie.ts
@@ -112,9 +112,19 @@ type ParseCookieOptions = {
   loose?: boolean | undefined
 }
 
+/**
+ * Parses a string into a Cookie object.
+ * @param str Cookie string to parse
+ * @returns `Cookie` object for valid string inputs, `undefined` for invalid string inputs,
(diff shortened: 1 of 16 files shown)


lib/cookie/cookie.ts
```diff
@@ -112,9 +112,19 @@ type ParseCookieOptions = {
...
+/**
+ * Parses a string into a Cookie object.
+ * @param str Cookie string to parse
+ * @returns `Cookie` object for valid string inputs, `undefined` for invalid string inputs,
+ * or `null` for non-string inputs or empty string
+ */
 function parse(
   str: string,
   options?: ParseCookieOptions,
+  // TBD: Should we change the API to have a single "invalid input" return type? I think `undefined`
```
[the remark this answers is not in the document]
colincasey: After browsing through this code some more, I'm leaning more towards we should do this now.

colincasey (changes requested):

lib/cookie/cookie.ts
```diff
@@ -387,18 +397,18 @@ const cookieDefaults = {
 type CreateCookieOptions = {
   key?: string
   value?: string
-  expires?: Date | 'Infinity' | null
+  expires?: Nullable<Date | 'Infinity'>
```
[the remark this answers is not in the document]
wjhsf: Yeah, the changes from the other PR will take precedence, since they make the types a bit tighter.

lib/memstore.ts
```diff
@@ -54,20 +59,20 @@ export class MemoryCookieStore extends Store {
...
 
   override findCookie(
-    domain: string | null,
-    path: string | null,
-    key: string | undefined,
-  ): Promise<Cookie | undefined>
+    domain: Nullable<string>,
+    path: Nullable<string>,
+    key: Nullable<string>,
+  ): Promise<Nullable<Cookie>>
```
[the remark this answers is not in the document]
wjhsf: Yeah, I think `T | undefined` is a reasonable return type. There are a few helper functions like `canonicalDomain` that return `T | null`. Should we change those as well? They're top-level package exports, so it's _technically_ a breaking change, but they're not the things people actually use this package for.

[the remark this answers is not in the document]
colincasey: I think so. I only ported them over like that to maintain parity with the `v4` code but the return value inconsistencies have always bothered me 😅

colincasey (approved):

lib/cookie/cookie.ts
```diff
@@ -276,6 +280,10 @@ function parse(
   return c
 }
 
+// TBD: `null` is valid JSON, but `undefined` is not. Options:
+// 1. Change this to return `undefined` - weird because it's not JSON
+// 2. Keep this as `null` - weird because it's a violation of our new convention
+// 3. Change *everything* to return `null` - a lot of work, maybe other edge cases that we can't change?
```
wjhsf, line 286: @colincasey thoughts on this?

test/parsing_test.js
```diff
@@ -245,12 +241,12 @@ vows
         return Cookie.parse("\x08", true) || null;
       },
       "doesn't parse": function(c) {
-        assert.equal(c, null);
+        assert.equal(c, undefined);
```
colincasey, line 244: You missed the `Cookie.parse(...) || null` above which means the value of `c` here is actually `null` but the test still passes since `assert.equal` is used instead of `assert.strictEqual`.

```suggestion
        assert.equal(c, SHOULD_BE_UNDEFINED);
```

test/parsing_test.js
```diff
@@ -533,10 +529,10 @@ vows
         };
       },
       "large one doesn't parse": function(c) {
-        assert.equal(c.cookie1, null);
+        assert.equal(c.cookie1, undefined);
```
colincasey, line 532: To be absolutely certain the value is `undefined`.
```suggestion
        assert.strictEqual(c.cookie1, undefined);
```

test/parsing_test.js
```diff
@@ -533,10 +529,10 @@ vows
...
       },
       "small one doesn't parse": function(c) {
-        assert.equal(c.cookie2, null);
+        assert.equal(c.cookie2, undefined);
```
colincasey, line 535: To be absolutely certain the value is undefined.
```suggestion
        assert.strictEqual(c.cookie2, undefined);
```

test/parsing_test.js
```diff
@@ -245,12 +241,12 @@ vows
         return Cookie.parse("\x08", true) || null;
```
wjhsf, line 241:
```suggestion
        return Cookie.parse("\x08", true) ?? SHOULD_BE_UNDEFINED;
```

colincasey (approved):

Issue #378, opened 2024-03-07
Why sameSite=None is not added for toString method


stefanbeigel:
https://github.com/salesforce/tough-cookie/blob/master/lib/cookie/cookie.ts#L612

As far as I know chrome treat cookies as SameSite=Lax by default if no SameSite attribute is specified.
So setting SameSite=None is not same as setting no SameSite.

So I wonder why it is returned by the toString method.

colincasey:
Good question @stefanbeigel. The `SameSite` attribute is not part of [rfc6265](https://datatracker.ietf.org/doc/html/rfc6265) and instead comes from the active draft [rfc6265bis](https://datatracker.ietf.org/doc/draft-ietf-httpbis-rfc6265bis/). The `SameSite` implementation for `tough-cookie` was added in [this commit](https://github.com/salesforce/tough-cookie/commit/a4bf3e265204359c351f173a30e6295d8a369e60) which references  [version 2 of httpbis-rfc6265bis](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-02#section-5.3.7):

> If the attribute-name case-insensitively matches the string
>   "SameSite", the user agent MUST process the cookie-av as follows:
>
>   1.  If cookie-av's attribute-value is not a case-insensitive match
>       for "Strict" or "Lax", ignore the "cookie-av".
>
>   2.  Let "enforcement" be "Lax" if cookie-av's attribute-value is a
>       case-insensitive match for "Lax", and "Strict" otherwise.
>
>   3.  Append an attribute to the cookie-attribute-list with an
>       attribute-name of "SameSite" and an attribute-value of
>       "enforcement".

But the steps above no longer match up with the [latest version of rfc6265bis](https://www.ietf.org/archive/id/draft-ietf-httpbis-rfc6265bis-13.html#section-5.5.7):

> Let enforcement be "Default".
> 
> If cookie-av's attribute-value is a case-insensitive match for "None", set enforcement to "None".
> 
> If cookie-av's attribute-value is a case-insensitive match for "Strict", set enforcement to "Strict".
> 
> If cookie-av's attribute-value is a case-insensitive match for "Lax", set enforcement to "Lax".
> 
> Append an attribute to the cookie-attribute-list with an attribute-name of "SameSite" and an attribute-value of enforcement.

The current implementation should be corrected to match the text above.

[...]

PR #381, opened 2024-03-13
Clean up cookie creation

wjhsf:
Currently, the props on `Cookie` all have `| undefined`, even though none of them actually ever use `undefined` as a value. If we get rid of `| undefined`, though, we get a TypeScript error complaining that the prop is not definitively assigned in the constructor. This is because we use `Object.assign` to initialize props, and I guess TypeScript just can't quite handle that. So, to get rid of `undefined` as a false possible value, we have to manually assign every prop. A bit more verbose, but not a significant change.

Additionally, the cookie constructor options are identical to the props of `Cookie`, excluding the methods, so I changed the type from explicitly listing everything to just a mapped type, for simpler maintenance. I wasn't sure if that would be a good idea, because sometimes mapped types get mangled, but the result type is still readable, so I think it's fine.

screenshot of the computed `CreateCookieOptions` type, showing that it is reported as a simple, readable key/value object type, rather than a hard-to-understand layering of mapped types

nb: This was originally part of #377, but has been split out for easier review.

### DIFF
--- a/lib/cookie/cookie.ts
+++ b/lib/cookie/cookie.ts
@@ -403,25 +403,39 @@ type CreateCookieOptions = {
 }
 
 export class Cookie {
-  key: string | undefined
-  value: string | undefined
-  expires: Date | 'Infinity' | null | undefined
-  maxAge: number | 'Infinity' | '-Infinity' | undefined
-  domain: string | null | undefined
-  path: string | null | undefined
-  secure: boolean | undefined
-  httpOnly: boolean | undefined
-  extensions: string[] | null | undefined
+  key: string
+  value: string
+  expires: Date | 'Infinity' | null
+  maxAge: number | 'Infinity' | '-Infinity' | null
+  domain: string | null
+  path: string | null
+  secure: boolean
+  httpOnly: boolean
+  extensions: string[] | null
   creation: Date | 'Infinity' | null
-  creationIndex: number | undefined
-  hostOnly: boolean | null | undefined
-  pathIsDefault: boolean | null | undefined
-  lastAccessed: Date | 'Infinity' | null | undefined
+  creationIndex: number
+  hostOnly: boolean | null
+  pathIsDefault: boolean | null
+  lastAccessed: Date | 'Infinity' | null
   sameSite: string | undefined
 
   constructor(options: CreateCookieOptions = {}) {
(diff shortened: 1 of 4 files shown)


lib/cookie/cookie.ts
```diff
@@ -382,46 +393,42 @@ const cookieDefaults = {
...
-  httpOnly?: boolean
-  extensions?: string[] | null
-  creation?: Date | 'Infinity' | null
-  creationIndex?: number
-  hostOnly?: boolean | null
-  pathIsDefault?: boolean | null
-  lastAccessed?: Date | 'Infinity' | null
-  sameSite?: string | undefined
-}
+} as const satisfies Required<CreateCookieOptions>
```
wjhsf, line 396: * `as const` makes the defaults a readonly value and uses the most precise type possible, e.g .`""` instead of `string`
* `satisfies Require<CreateCookieOptions>` ensures that we adhere to that interface without losing the precision of the type we already have, e.g. we keep `"Infinity"` for `expires`, rather than getting `Date | "Infinity" | null` from `CreateCookieOptions`

colincasey (approved):

lib/cookie/cookie.ts
```diff
@@ -382,46 +393,42 @@ const cookieDefaults = {
...
   creation: Date | 'Infinity' | null
-  creationIndex: number | undefined
-  hostOnly: boolean | null | undefined
-  pathIsDefault: boolean | null | undefined
-  lastAccessed: Date | 'Infinity' | null | undefined
+  creationIndex: number
+  hostOnly: boolean | null
+  pathIsDefault: boolean | null
+  lastAccessed: Date | 'Infinity' | null
   sameSite: string | undefined
```
wjhsf, line 413: @colincasey `sameSite` uses `undefined` as a real value, so I left it for this PR as changing it to `null` would break things. Do you think, as a follow up, we should change it to use `null` instead? It's weird to have the inconsistency, but changing it would be a breaking change.

colincasey: Let's leave it like this for now.  I think we'll have to revisit `sameSite` anyways after some investigation prompted by https://github.com/salesforce/tough-cookie/issues/378 showed we're not fully compliant with the latest guidance for that attribute.

PR #383, opened 2024-03-15
Enforce explicit function return type

wjhsf:
The primary motivation is to help with #380; enforcing an explicit return type helps us identify which functions return `null` / `undefined` and therefore standardize on one value as much as possible.

A secondary motivation is that it technically makes type checks faster, as parsing function bodies is no longer necessary to infer the return type. However, given the current size/state of our code, there's no discernible performance gain by enforcing this rule.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -13,6 +13,7 @@
   },
   "reportUnusedDisableDirectives": true,
   "rules": {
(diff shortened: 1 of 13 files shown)


colincasey (approved):
Good call to enable the explicit return type rule.

PR #384, opened 2024-03-17
Update v5 docs

colincasey:
This PR will be the staging area for multiple stacked PRs that update our API docs with markdown pages generated from TypeScript signatures + `@tsdoc` annotations. I'm introducing the [API Extractor](https://api-extractor.com/) into our develop toolchain to handle this.

The main landing page which is linked to from the README is here:
https://github.com/salesforce/tough-cookie/blob/v5-docs/api/docs/tough-cookie.md

And the README changes:
https://github.com/salesforce/tough-cookie/blob/v5-docs/README.md

The idea is to:
- keep API docs are accurate since they will be generated automatically to match code
- allow for our `README.md` to be simpler since API docs can be moved to a dedicated area
- provide some guardrails against breaking API changes using the API Extractor tooling since the API Report output can be used as a basis for [API review workflow](https://api-extractor.com/pages/overview/demo_api_report/)

--- 

> [!NOTE]
> This PR will be marked as a draft until all the following child PRs in this stack have been reviewed and merged
- #385 
- #386 
- #387 
- #388 
- #390 
- #391 
- #393
- #394 
- #395 
- #396 
- #397 
- #398 
- #399 
- #400 
- #401 
- #402 
- #407 
- #408


lib/cookie/cookie.ts
```diff
@@ -401,23 +426,101 @@ const cookieDefaults = {
   sameSite: undefined,
 } as const satisfies Required<CreateCookieOptions>
 
+/**
+ * An HTTP cookie (web cookie, browser cookie) is a small piece of data that a server sends to a user's web browser.
+ * It is defined in {@link https://www.rfc-editor.org/rfc/rfc6265.html | RFC6265}.
+ * @public
+ */
 export class Cookie {
```
wjhsf, line 434:
```suggestion
export class Cookie implements Required<CreateCookieOptions> {
```
`CreateCookieOptions` previously was computed from `Cookie`. Now that they are both declared independently, we could use `implements` to guard against accidentally diverging definitions.

wjhsf:
Minor TODO before final merge: Bump the `max-lines` limit in files that have way too many lines now (probably `cookie.ts` and `cookieJar.ts`).

lib/cookie/cookie.ts
colincasey: I'm not sure this is necessary. It would catch a new field added to `CreateCookieOptions` but it wouldn't catch a new field added to `Cookie` so these two can still get out of sync. Also, it stands out to me as a weird type signature if we're implementing a class against it's creation options. 

I believe the risk of these definitions getting out of sync is low since any code to support a new or renamed field would inevitably force us to modify both `Cookie` and `CreateCookieOptions` as well as other definitions.

But now that we have a solid foundation for TypeScript in this library I think we should look at revising our types to tackle some of these issues in our next major release.

wjhsf (dismissed):
lines changed: +13,344 -842

🚀

PR #392, opened 2024-03-21
Enable strict type checked rules

wjhsf:
When doing #383, I noticed that we extend `@typescript-eslint/recommended-type-checked` and `@typescript-eslint/strict`. There's also `@typescript-eslint/strict-type-checked` which is includes those two configs, plus a few additional type checked rules.

I also moved us from `.eslintrc` to the new flat config, just because we can!


package.json
```diff
@@ -92,31 +92,32 @@
     "test:legacy": "npm run build -- --declaration false && ./test/scripts/vows.js test/*_test.js",
     "typecheck": "tsc --noEmit",
     "cover": "jest --coverage",
-    "eslint": "eslint --env node --ext .ts .",
+    "lint": "eslint .",
```
wjhsf, line 95: Because every other project I work on uses `lint`, not `eslint`.

package-lock.json
colincasey, line 1: @wjhsf there are some nexus resolutions in this package.json that prevent it from being usable

colincasey (changes requested):
@wjhsf not sure what's going on with this PR but I tried to work around the bad resolutions in `package-lock.json` by just deleting and regenerating it locally with a fresh `npm install`. But, after doing that, I get errors when I execute `npm run lint` that `eslint-config-prettier` cannot be found.

package-lock.json
wjhsf: 🫠

colincasey:
@wjhsf I'm not so sure about these changes. If you check the test output of https://github.com/salesforce/tough-cookie/actions/runs/8392570051/job/22985505761?pr=392 it looks like there are regressions introduced with the vows test suite which we're using as a sanity check for these types of refactorings. 

Maybe a better approach is to disable all the new rules introduced with `@typescript-eslint/strict-type-checked` and have incremental PRs with each rule enabled + it's required changes?

wjhsf:
Since we don't super care about how clean the vows tests are, I've just disabled the violated rules instead of fixing them, except one violation that's trivial.

eslint.config.mjs
```diff
@@ -0,0 +1,43 @@
...
+      parserOptions: {
+        project: './tsconfig.json',
+        tsconfigDirName: import.meta.dirname,
+      },
+    },
+    linterOptions: {
+      reportUnusedDisableDirectives: 'warn',
+    },
+    rules: {
+      'max-lines': ['warn', 500],
```
colincasey, line 27: I ended up bumping this to `2000` after the tsdocs where introduced in #384. Should we just disable this rule entirely?

 ```suggestion
      'max-lines': ['warn', 2000],
```

colincasey (approved):

PR #395, opened 2024-03-23
V5 Docs - `parseDate`

colincasey:
Rendered pages:
- https://github.com/salesforce/tough-cookie/blob/v5-docs-parsedate/api/docs/tough-cookie.parsedate.md

---
Main PR:
- #384


api/docs/tough-cookie.parsedate.md
```diff
@@ -0,0 +1,122 @@
...
+
+</td></tr>
+</tbody></table>
+**Returns:**
+
+Date \| undefined
+
+## Remarks
+
+\#\#\# RFC6265 - 5.1.1. Dates
```
wjhsf, line 56: This isn't rendered properly 🫤

wjhsf (approved):

api/docs/tough-cookie.parsedate.md
colincasey: I'll file an issue about the generated content against `@microsoft/api-documenter` if one doesn't already exist.

PR #397, opened 2024-03-24
V5 Docs - `permutePath`

colincasey:
Rendered pages:
- https://github.com/salesforce/tough-cookie/blob/v5-docs-permutepath/api/docs/tough-cookie.permutepath.md

---
Main PR:
- #384


api/docs/tough-cookie.permutepath.md
```diff
@@ -0,0 +1,61 @@
...
+</td><td>
+
+the path to generate permutations for
+
+
+</td></tr>
+</tbody></table>
+**Returns:**
+
+string\[\]
```
wjhsf, line 52: This doesn't seem to be formatted correctly. 🫤

api/docs/tough-cookie.permutepath.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Example
+
+
+```
```
wjhsf, line 57: Examples seem to be missing the `ts` language tag for syntax highlighting. Is there any way to add it?

wjhsf (approved):

api/docs/tough-cookie.permutepath.md
colincasey: I noticed this too and will file a bug with `@microsoft/api-documenter` if one doesn't already exist.

api/docs/tough-cookie.permutepath.md
colincasey: Yes, it looks like adding `ts` to the block is respected in the output. 

I believe I missed this for several of the PRs. Once these are all merged into the `v5-docs` branch I'll need to do a final bit of cleanup (e.g.; removing the doc `.gitignore`, verifying there are no warnings about missing types from our public api, etc.) so I'll add the following to the list:
- annotate examples with `ts` for syntax highlighting

PR #398, opened 2024-03-24
V5 Docs - `PrefixSecurity`

colincasey:
Rendered pages:
- https://github.com/salesforce/tough-cookie/blob/v5-docs-prefixsecurity/api/docs/tough-cookie.prefixsecurity.md

---
Main PR:
- #384


api/tough-cookie.api.md
```diff
@@ -264,38 +266,38 @@ export class ParameterError extends Error {
...
 export function permutePath(path: string): string[];
 
-// @public (undocumented)
-export const PrefixSecurityEnum: Readonly<{
-    SILENT: "silent";
-    STRICT: "strict";
-    DISABLED: "unsafe-disabled";
-}>;
+// @public
+export enum PrefixSecurity {
```
wjhsf, line 284: Enums are gross and it's a best practice to avoid them. What does the documentation look like if we use a readonly object?

colincasey: The field-level docs don't render in that case.  You get a page with just `export const PrefixSecurityEnum` which is kind of useless but maybe this should be explored further.  

I also think I should revert this rename since this export exists in `v4` (even though an `enum` with the suffix `Enum` offends me 😅)

colincasey: @wjhsf reverted this `enum` and naming change so the previous `PrefixSecurityEnum` as a readonly constant is retained. 

The property docs are not supported when doing this so I've moved each property description into the main description as bullet points.

wjhsf (approved):

PR #400, opened 2024-03-26
V5 Docs - `Cookie`

colincasey:
Rendered pages:
- https://github.com/salesforce/tough-cookie/blob/v5-docs-cookie/api/docs/tough-cookie.cookie.md
- https://github.com/salesforce/tough-cookie/blob/v5-docs-cookie/api/docs/tough-cookie.createcookieoptions.md
- https://github.com/salesforce/tough-cookie/blob/v5-docs-cookie/api/docs/tough-cookie.parsecookieoptions.md
- https://github.com/salesforce/tough-cookie/blob/v5-docs-cookie/api/docs/tough-cookie.serializedcookie.md

---
Main PR: 
- #384


api/docs/tough-cookie.createcookieoptions.md
```diff
@@ -0,0 +1,17 @@
...
+## CreateCookieOptions type
+
+Type alias for all the cookie properties defined by [Cookie](./tough-cookie.cookie.md) (except for [Cookie.creationIndex](./tough-cookie.cookie.creationindex.md)<!-- -->).
+
+**Signature:**
+
+```typescript
+export type CreateCookieOptions = Omit<{
+    [K in keyof Cookie as Cookie[K] extends (...args: any[]) => any ? never : K]?: Cookie[K];
+}, 'creationIndex'>;
```
wjhsf, line 14: This is not helpful. Can we tell the docs generator to print the _result_ type, rather than the source? Otherwise we might just want to make the interface static/explicit.

wjhsf (approved):

api/docs/tough-cookie.createcookieoptions.md
colincasey: I agree, not helpful at all.  I don't think the doc generator is smart enough to do this but it warrants a closer look.

colincasey: @wjhsf it doesn't seem to be possible to do this in TSDoc without the assistance of a custom plugin to expand the type information. I've changed the above to an explicit interface that lists all the properties as optional and used `@inheritDoc` on each property to reuse the existing documentation from the Cookie properties.

PR #402, opened 2024-03-29
V5 Docs - `Store` and `MemoryStore`

colincasey:
Rendered pages:
- https://github.com/salesforce/tough-cookie/blob/v5-docs-stores/api/docs/tough-cookie.store.md
- https://github.com/salesforce/tough-cookie/blob/v5-docs-stores/api/docs/tough-cookie.memorycookiestore.md

---
Main PR:
- #384


api/docs/tough-cookie.store.md
```diff
@@ -0,0 +1,351 @@
...
+## Store class
+
+Base class for [CookieJar](./tough-cookie.cookiejar.md) stores.
+
+The storage model for each [CookieJar](./tough-cookie.cookiejar.md) instance can be replaced with a custom implementation. The default is [MemoryCookieStore](./tough-cookie.memorycookiestore.md)<!-- -->.
+
+**Signature:**
+
+```typescript
+export declare class Store
```
wjhsf, line 14: This doesn't seem all that helpful. 😅

api/docs/tough-cookie.memorycookiestore.md
```diff
@@ -0,0 +1,344 @@
...
+
+</th><th>
+
+Description
+
+
+</th></tr></thead>
+<tbody><tr><td>
+
+[(constructor)()](./tough-cookie.memorycookiestore._constructor_.md)
```
wjhsf, line 36: `(constructor)` is kinda weird - is it a convention from the doc generator?

lib/memstore.ts
```diff
@@ -48,27 +17,66 @@ export type MemoryCookieStoreIndex = {
...
     this.synchronous = true
     this.idx = Object.create(null) as MemoryCookieStoreIndex
   }
 
+  /**
+   * Retrieve a {@link Cookie} with the given `domain`, `path`, and `key` (`name`). The RFC maintains that exactly
+   * one of these cookies should exist in a store. If the store is using versioning, this means that the latest or
+   * newest such cookie should be returned.
+   *
+   * Callback takes an error and the resulting Cookie object. If no cookie is found then null MUST be passed instead (that is, not an error).
```
wjhsf, line 50:
```suggestion
```
No callback in this signature!

image

lib/memstore.ts
```diff
@@ -48,27 +17,66 @@ export type MemoryCookieStoreIndex = {
   }
 }
 
+/**
+ * An in-memory {@link Store} implementation for {@link CookieJar}. This is the default implementation used by
+ * {@link CookieJar} and supports both async and sync operations. Also supports serialization, getAllCookies, and removeAllCookies.
```
wjhsf, line 22: Should we warn "this is mostly a development tool, you should use something better in production"?

lib/memstore.ts
```diff
@@ -48,27 +17,66 @@ export type MemoryCookieStoreIndex = {
...
+ * @public
+ */
 export class MemoryCookieStore extends Store {
+  /**
+   * This value is `true` since {@link MemoryCookieStore} implements synchronous functionality.
+   */
   override synchronous: boolean
```
wjhsf, line 29:
```suggestion
  override synchronous: true
```
Not super important, but does this work? 🤷

api/docs/tough-cookie.store.md
colincasey: Perhaps, but it is accurate 😄

lib/memstore.ts
colincasey: Does that even need to be stated? Also, there are some cases where you might have a use for an in-memory store in production. For example, if you are an automation tool that needs to login to a site and store an auth cookie.

lib/memstore.ts
colincasey: No, it breaks the tests.

PR #408, opened 2024-04-11
V5 Docs - GitHub Action Changes

colincasey:
This builds on top of #384 to add CI checks that validate and fail the build if:
* Code changes from the PR would alter the public API in any way
* Generated docs are not committed

If either of the above fails, it means the whoever opened the PR needs to execute `npm run api:dev` to rebuild the API report + generated docs and commit them with the PR. This is beneficial since we'll see when changes will affect the public API or documentation.


awaterma (approved):
Approved

wjhsf (approved):

PR #421, opened 2024-06-03
chore: update deps

wjhsf:
Dependabot PRs seem borked, so here's a manual bump.


colincasey:
@wjhsf we may want to create a [dependabot group](https://github.blog/changelog/2023-06-30-grouped-version-updates-for-dependabot-public-beta/) for eslint. That library and it's plugins seem to need extra care with upgrading especially when TypeScript is involved.

colincasey (approved):

PR #423, opened 2024-06-21
Use latest TypeScript version

wjhsf:
Now that we're a TypeScript project, changing TypeScript version is a breaking change. We only get one TypeScript version per major release, so we should use the latest version, rather than the version that was current when we started moving to TypeScript ages ago.

Code change is a linter fix, I guess due to better type detection.

### DIFF
--- a/api/tough-cookie.api.md
+++ b/api/tough-cookie.api.md
@@ -4,8 +4,6 @@
 
 ```ts
 
-/// <reference types="node" />
-
 // @public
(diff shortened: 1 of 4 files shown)


colincasey (dismissed):
Yeah, when we first discussed this TypeScript v5 was just released. With v5 usage being double that of v4 right now I think it makes sense to switch over.

colincasey (approved):

PR #424, opened 2024-06-23
Fix validation errors not calling callbacks

colincasey:
I noticed this issue while upgrading `tough-cookie` in `jsdom` which was causing their test suite to hang. There is a sneaky little bug around our internal validations where, in several methods that accept callbacks, our validations error out directly and never invoke the given callback with the error. This can lead to code that waits for a callback that will never complete. This also affects functions wrapped in `Sync` which leverage callbacks to capture their return values.

The affected methods were:
- `getCookiesSync`
- `getCookieString`
- `getSetCookieStrings`
- `serializeSync`
- `serialize`
- `setCookieSync`
- `setCookie`
- `getCookiesSync`
- `getCookies`

There were also several validation errors hiding in support methods. These appear to be redundant since the type signatures present already do much of the validation. These have been removed from:
- `checkSameSiteContext`
- `isHostPrefixConditionMet`
- `isSecurePrefixConditionMet`
- `cookieCompare`
- `formatDate`
- `permutePath`


lib/cookie/cookieJar.ts
```diff
@@ -330,12 +327,22 @@ export class CookieJar {
...
+      })
+    } catch (err) {
+      if (err instanceof Error) {
+        syncErr = err
+      } else {
+        throw err
+      }
+    }
+
     // These seem to be false positives; it can't detect that the value may be changed in the callback
```
wjhsf, line 344:
```suggestion
```
Comment is for the removed `// eslint-disable`

lib/cookie/cookieJar.ts
```diff
@@ -330,12 +327,22 @@ export class CookieJar {
...
+      fn.call(this, (error: Error | null, result?: T | undefined) => {
+        syncErr = error
+        syncResult = result
+      })
+    } catch (err) {
+      if (err instanceof Error) {
+        syncErr = err
+      } else {
+        throw err
+      }
```
wjhsf, line 341: Elsewhere we just cast `err as Error`, why be particular here?

wjhsf (approved):

lib/cookie/cookieJar.ts
colincasey: Fixed

lib/cookie/cookieJar.ts
colincasey: Changed to cast.

wjhsf (approved):

PR #431, opened 2024-07-18
Checks structure instead of instanceof for URL test

colincasey:
This is a follow up into my testing of our latest release candidate against the `jsdom` test suite. The callback issues have been sorted out but I also encountered a validation issue around the `URL` object that is used by `jsdom`. When we are passed a `URL` object, we test it using `url instanceof URL` but the `URL` object from `jsdom` is not the same as the global `URL` object. 

Relying on `instanceof` is always tricky in JavaScript. We should instead favor checking for the structure of the object we expect which must have the following structure:

```ts
type UrlContext = {
  hostname: string
  pathname: string
  protocol: string
}
```

This PR replaces the `instanceof` check with a validation for the above structure. This will pass for unknown `url` values that are instances of `URL` since they will have that structure and is already represented by the following test:
https://github.com/salesforce/tough-cookie/blob/master/lib/__tests__/cookieJar.spec.ts#L1252-L1265

### DIFF
--- a/lib/cookie/cookieJar.ts
+++ b/lib/cookie/cookieJar.ts
@@ -188,9 +188,28 @@ export interface CreateCookieJarOptions {
 const SAME_SITE_CONTEXT_VAL_ERR =
   'Invalid sameSiteContext option for getCookies(); expected one of "strict", "lax", or "none"'
 
-function getCookieContext(url: unknown): URL {
-  if (url instanceof URL) {
-    return url
+type UrlContext = {
+  hostname: string
+  pathname: string
+  protocol: string
+}
+
+function getCookieContext(url: unknown): UrlContext {
+  if (
+    url &&
+    typeof url === 'object' &&
+    'hostname' in url &&
+    typeof url.hostname === 'string' &&
+    'pathname' in url &&
+    typeof url.pathname === 'string' &&
+    'protocol' in url &&
+    typeof url.protocol === 'string'
+  ) {
+    return {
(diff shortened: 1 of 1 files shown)


lib/cookie/cookieJar.ts
```diff
@@ -188,9 +188,28 @@ export interface CreateCookieJarOptions {
...
+function getCookieContext(url: unknown): UrlContext {
+  if (
+    url &&
+    typeof url === 'object' &&
+    'hostname' in url &&
+    typeof url.hostname === 'string' &&
+    'pathname' in url &&
+    typeof url.pathname === 'string' &&
+    'protocol' in url &&
+    typeof url.protocol === 'string'
```
wjhsf, line 206: Could be nice to extract this to a `isUrlLike` helper. I also feel like we should skip the `in` checks and just do `typeof` checks. And tell TypeScript to shush, because we're confident in what we're doing here.

wjhsf (approved):

lib/cookie/cookieJar.ts
colincasey: I don't mind the extra checks seeing as how TypeScript does require them since the object is `unknown`. I'm more confident that the TypeScript compiler knows what it's doing than me most days 😅.

Also, the `getCookieContext` is essentially already a `isUrlLike` helper. The naming is just odd since we're keeping a lot of the legacy function names untouched. Since we don't duplicate this code anwhere else, I don't think it needs to be extracted.

PR #433, opened 2024-07-22
use `domainToASCII(str)` instead of `new URL(str).hostName`

wjhsf:
Today I learned that node has a utility, [`domainToASCII`](https://nodejs.org/api/url.html#urldomaintoasciidomain), that converts a domain name to ASCII. We can use that in `canonicalDomain` because we don't need to parse the full URL.


colincasey (approved):

sam3d:
This PR sadly prevents `tough-cookie` from being used in the edge runtime because of `node:`.

colincasey:
Thanks for pointing that out @sam3d. I've created two issues, #479 and #477, related to this.

PR #434, opened 2024-07-31
Enable isolated declarations

wjhsf:
I recently learned about [isolated declarations](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-5.html#isolated-declarations), which enforces some rules on type definitions that make it easier for downstream tooling to get type information without running a full type check. In order to do this, we have to slightly hide our custom inspect function.

I also switched us from `importsNotUsedAsValues` to `verbatimModuleSyntax` because the former is deprecated in favor of the latter.

### DIFF
--- a/lib/__tests__/data/dates/bsd-examples.ts
+++ b/lib/__tests__/data/dates/bsd-examples.ts
@@ -219,4 +219,4 @@ export default [
     test: 'IAintNoDateFool',
     expected: null,
   },
-]
+] as const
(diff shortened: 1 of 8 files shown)


wjhsf:
Our tooling for API documentation generation doesn't recognize `isolatedDeclarations`. Since this isn't a priority, I'm going to close for now; we can revisit once the tooling has been updated.

PR #435, opened 2024-07-31
use ESM instead of CJS

wjhsf:
Turns out we just forgot to specify `"type": "module"` in the `package.json`. This makes us a pure ESM project!


wjhsf:
vows is the only thing blocking this change; we want to keep vows for the v5 release, but we'll drop it for v6.

wjhsf:
Closing in favor of #506.

PR #453, opened 2024-09-09
chore: streamline package publishing

wjhsf:
Following today's v5 release, we saw a few points of improvement on our release process:
1. The API docs include the current version number; docs are now updated as part of the `version` script.
2. Two tests use a hard-coded version of the project. There's not much value added by doing that, and it's annoying to maintain, so I changed them to use the version file.
  a. The version file is also something we've considered removing, but in the meantime I've added a test to make sure that the file just exports a valid semver string.
3. Added a GitHub Actions workflow to publish to npm, triggered by creating a GitHub release.

The new release process is:
1. Check out code to release
2. `npm version <type>`, using major/minor/etc. as appropriate
3. `git push && git push tags/<version>`
4. Create GitHub release
5. 🎉 it's on npm!

**Important:** Publishing will require creating/storing an `NPM_TOKEN` secret. So gotta do that before this will actually work.

This PR also contains a few prettier fixes because that happens sometimes.


package.json
```diff
@@ -92,20 +92,22 @@
     "!__tests__"
   ],
   "scripts": {
-    "api:dev": "npm run build && npm run api:extract -- --local && npm run api:docs",
-    "api:docs": "api-documenter markdown --input-folder ./tmp --output-folder ./api/docs",
-    "api:extract": "api-extractor run --verbose",
+    "docs": "npm run build && npm run docs:ci",
+    "docs:ci": "npm run docs:extract -- --local && npm run docs:generate",
+    "docs:generate": "api-documenter markdown --input-folder ./tmp --output-folder ./api/docs",
+    "docs:extract": "api-extractor run --verbose",
```
colincasey, line 98: I'm not sure I agree with this specific renaming. There are actually [two distinct processes](https://api-extractor.com/pages/overview/intro/) happening here:
1. Public API metadata is extracted from the code 
2. Documentation is generated from the code 

If we're going to rename, we should probably use names that reflect these different processes so we don't conflate them as I originally did.

Also, we shouldn't use pass `--local` in the `docs:ci` script. That flag disables the validation which checks if the public API has changed. The API report that `api-extractor` produces should be thought of as similar to `package-lock.json` only it's a lockfile for public API changes, not dependencies. Following that comparison, consider the workflow for `package-lock.json`:
- Changes are deliberately made to the lockfile by a developer (or dependabot) locally and then the change is committed
- When running `npm ci` in an automated environment, if the lockfile doesn't match `package.json` then it will error

We should follow a similar process for the API report. That is how the previous scripts operated with `api:dev` being the deliberate, local change and `api:extract` being the automated one which should fail if there is a difference detected.

colincasey (changes requested):

colincasey (approved):

PR #457, opened 2024-09-16
fix: remove cookies that expire at epoch time of `0`

### LINKED ISSUE #455: Cookies with expiry time set to epoch zero are matched and not expired
If the server sends a cookie with expiry time set to epoch zero (1970-01-01T00:00:00.000Z), the cookie is then returned in ``getCookies(url)`` even though "expire" is set to true (the default)

This is because ``expiryTime`` here should be checked for ``undefined`` not just ``!expiryTime``
https://github.com/salesforce/tough-cookie/blob/1a71340594beac95c4ea10fefe4eb79fdfead5c2/lib/cookie/cookieJar.ts#L940

colincasey:
The logic for checking and removing expired cookies fails when the cookie has an expiry date set to the epoch time of `0` (i.e.; `Thu, 01 Jan 1970 00:00:00 GMT`). This is because the branch check that triggers cookie removal was testing `expiryTime` which is `falsy` for a value of `0` and that short-circuits the logic.

This PR updates the logic to narrow the type of `expiryTime` from `number | undefined` to `number` using `expiryTime != undefined` which allows the `0` value to proceed to the subsequent condition and trigger the cookie removal.

Fixes #455

### DIFF
--- a/lib/__tests__/cookieJar.spec.ts
+++ b/lib/__tests__/cookieJar.spec.ts
@@ -1521,6 +1521,16 @@ describe('validation errors invoke callbacks', () => {
   })
 })
 
+it('issue #455 - should expire a cookie with epoch zero', async () => {
+  const cookieJar = new CookieJar()
+  await cookieJar.setCookie(
+    'OptionsTest=FooBar; Expires=Thu, 01 Jan 1970 00:00:00 GMT;',
+    'http://example.com',
+  )
+  const cookies = await cookieJar.getCookies('http://example.com')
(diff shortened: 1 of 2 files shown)


EZaykov:
Iam not sure but it seems like this fix did not showed up (at least in compiled js).

In node_modules its still `if (expireCheck && expiryTime && expiryTime <= now)`.

Made small patch in my store:

colincasey:
@EZaykov this fix should be in the latest release [v5.1.0](https://github.com/salesforce/tough-cookie/discussions/485)

PR #459, opened 2024-09-25
Restore missing `expiryDate` method

### LINKED ISSUE #458: `expiryDate` method missing in v5.0.0
The `expiryDate` method on the `Cookie` class was accidentally removed in the v5 release. It will be restored in the v5.0.1 release.

colincasey:
Fixes #458


lib/cookie/cookie.ts
```diff
@@ -856,6 +856,35 @@ export class Cookie {
...
+    const millisec = this.expiryTime(now)
+    if (millisec == Infinity) {
+      // The 31-bit value of 2147483647000 was chosen to be the MAX_TIME representable
+      // in tough-cookie though MDN states that the actual maximum value for a Date is 8.64e15.
+      // I'm guessing this is due to the Y2038 problem that would affect systems that store
+      // unix time as 32-bit integers.
+      // See:
+      // - https://github.com/salesforce/tough-cookie/commit/0616f70bf725e00c63d442544ad230c4f8b23357
+      // - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#the_epoch_timestamps_and_invalid_date
+      // - https://en.wikipedia.org/wiki/Year_2038_problem
```
wjhsf, line 879: (aside) I was curious whether there was anything in the cookie spec to this effect. Turns out that Chrome has a [cap of 400 days](https://developer.chrome.com/blog/cookie-max-age-expires/), and they've proposed [adding it to the spec](https://httpwg.org/http-extensions/draft-ietf-httpbis-rfc6265bis.html#name-cookie-lifetime-limits).

lib/__tests__/cookieJar.spec.ts
```diff
@@ -1247,6 +1248,9 @@ it('should fix issue #154 - Expiry should not be affected by creation date', asy
   ])
   // the expiry time should be 60s from now (1000)
   expect(updatedCookies[0]?.expiryTime()).toBe(now + 60 * 1000 + 1000)
+  expect(updatedCookies[0]?.expiryDate()).toEqual(
+    new Date(now + 60 * 1000 + 1000),
+  )
```
wjhsf, line 1253: At some point we should probably go through and ensure we actually assert the existence of everything we expect to have. 😅

wjhsf (approved):

lib/cookie/cookie.ts
colincasey: Perfect. I didn't like the look of this arbitrary date and would prefer having it be defined in the spec.

lib/__tests__/cookieJar.spec.ts
colincasey: Yeah, we also were operating under the assumption that the vows test suite would be our safety net to prevent us from making incorrect changes to the public API. As we've seen, that can be misleading if there is no coverage for a method.

PR #487, opened 2025-01-09
chore: avoid nodejs modules

wjhsf:
This adds a new eslint plugin to avoid node's built in modules, and hopefully help keep our package more cross-platform compatible.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -2,19 +2,17 @@
 
 import eslint from '@eslint/js'
 import prettierRecommended from 'eslint-plugin-prettier/recommended'
-// eslint-disable-next-line import/no-unresolved
-import tseslint from 'typescript-eslint'
+import { config, configs } from 'typescript-eslint'
 import { flatConfigs as pluginImport } from 'eslint-plugin-import'
 import globals from 'globals'
 
-export default tseslint.config(
+export default config(
   {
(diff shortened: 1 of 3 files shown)


eslint.config.mjs
```diff
@@ -2,7 +2,9 @@
 
 import eslint from '@eslint/js'
 import prettierRecommended from 'eslint-plugin-prettier/recommended'
+// eslint-disable-next-line import/no-unresolved
```
colincasey, line 5: I think this is due to https://github.com/import-js/eslint-plugin-import/issues/1485 which includes a fix that works for me (locally) but produces a couple extra warnings.

eslint.config.mjs
```diff
@@ -11,6 +13,9 @@ export default tseslint.config(
   },
   eslint.configs.recommended,
   ...tseslint.configs.strictTypeChecked,
+  pluginImport.recommended,
+  // @ts-expect-error not sure why the plugin isn't type correctly, but it works...
```
colincasey, line 17: I removed this after including the fix mentioned above and did not see any errors.

colincasey:
Nice! This was the plugin I was looking at last night when trying to figure out how we can best avoid breaking non-Node runtimes.

Scutua (dismissed):

colincasey (approved):

PR #492, opened 2025-02-04
Prototype pollution patch

reefgb96:


wjhsf:
The prototype pollution vulnerability exposed via `MemoryCookieStore` has been patched since [version 4.1.3](https://github.com/salesforce/tough-cookie/releases/v4.1.3).

PR #493, opened 2025-02-04
chore: auto-close spam PRs

wjhsf:


.github/workflows/slonser.yaml
```diff
@@ -10,5 +10,5 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - env:
-          PR: "${{ github.event.number }}"
+          PR: '${{ github.event.number }}'
         run: gh pr close "$PR" --comment "The prototype pollution vulnerability exposed via `MemoryCookieStore` ([CVE-2023-26136](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2023-26136)) has been patched since [version 4.1.3](https://github.com/salesforce/tough-cookie/releases/v4.1.3). Please upgrade your code to use the latest version of `tough-cookie`."
```
colincasey, line 14: @wjhsf Can we also include steps to mitigate in the comment? E.g.;
- If you're doing this for `request` then it's safe to upgrade to v4.1.3 or higher
- Others should use resolutions/overrides or patch features from npm, pnpm, or Yarn

PR #495, opened 2025-02-06
Patch prototype pollution in tough-cookie@2.5.0 and add test for fix

yudavinchi:


wjhsf:
@yudavinchi why did you submit this PR?

wjhsf:
@yudavinchi was this part of some sort of coursework? What course was this for?

Issue #382, opened 2024-03-15
Secure; Flag should be ignored when sending to 'localhost'


ertl:
According to https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#restrict_access_to_cookies it is fine to send cookies via http even when the flag `secure` is set, as long as the receiving host is localhost.

Despite this, a previous attempt to address this problem through pull request #281 was made and subsequently closed. We need to reassess the decision to close the pull request and carefully consider the security implications of allowing cookies to be sent via HTTP in this scenario.

colincasey:
To be clear, the contributor of #281 self-closed their own pull request. No formal decision was made by the maintainers of this project regarding the validity of that PR 😅 

While nothing in RFC6265 explicitly states that there should be an exception for `localhost` and the `Secure` attribute, there is some text around secure channels in [section 4.1.2.5](https://www.rfc-editor.org/rfc/rfc6265.html#section-4.1.2.5)

> The Secure attribute limits the scope of the cookie to "secure" channels (where "secure" is defined by the user agent) 

This is an area where I think `tough-cookie` would benefit from having a better internal representation of a User Agent or User Agent Policy to assist in either determining or configuring how the library behaves in cases like this.

Can you give some more details around your use case @ertl?

ertl:
@colincasey Thank you for your response and for clarifying the status of the previous pull request.

You're right to point out that RFC6265 doesn't explicitly mention any exceptions for localhost and the Secure attribute. However, the text around secure channels in section 4.1.2.5 does leave room for interpretation regarding what constitutes a "secure" channel.

Our scenario includes an additional proxy app running on localhost. We believe that accessing the proxy via HTTPS is not practical because the connection through localhost is inherently secure due to its local nature, which limits external access and reduces the risk of interception or tampering.

colincasey:
@ertl that makes sense. We'd be open to a PR here if that's something you wanted to attempt. Otherwise, this will go in our list of v6 project goals.

Chriss4123:
Hey, I opened a PR for this issue [here](https://github.com/salesforce/tough-cookie/pull/498) which implements an RFC 6761 compliant URL trustworthiness check that allows `secure` cookies to work with `localhost` or loopback addresses.

PR #498, opened 2025-02-16
feat: Add RFC 6761–compliant localhost loopback checks so secure cookies work on localhost (fixes: #382)

Chriss4123:
fixes: [#382](https://github.com/salesforce/tough-cookie/issues/382)

# Description

This commit extends how cookies are treated in secure contexts by fully recognizing `localhost` and loopback IPs as trustworthy origins, matching the de facto behavior of all modern browsers and RFC 6761. Previously, `tough-cookie` defaulted `secure` to `true` only for `https:` and `wss:` URLs, causing cookies with `secure` to not work with `localhost`.

### What Changed

1. **New `secureContext.ts`**  
   - Implements `isPotentiallyTrustworthy(url)` by checking:
     - Schemes: `https` or `wss`
     - Loopback IPs: `127.0.0.1/8` and `::1`  
     - Hostnames: `localhost` and `*.localhost`  
   - Adapted from Chromium’s `IsLocalhost`, `IsLoopback` and `HostNoBracketsPiece`, located at:
     - [`IsLocalhost`](https://github.com/chromium/chromium/blob/main/net/base/url_util.cc#L457-L551)
     - [`IsLoopback`](https://github.com/chromium/chromium/blob/main/net/base/ip_address.cc#L228-L243)
     - [`HostNoBracketsPiece`](https://github.com/chromium/chromium/blob/main/url/gurl.cc#L440-L448)

2. **`cookieJar.ts` Update**
  - Replaced the old check which only accounts for `https` and `wss` schemes:
```ts
const secure =
  context.protocol &&
  (context.protocol == 'https:' || context.protocol == 'wss:')
```
   - with the new `isPotentiallyTrustworthy` function which accounts for secure schemes, `localhost` addresses and loopback addresses.

No existing functionality is removed. Where the old check would return true, so does the new check. The only added functionality is treating `localhost` addresses and loopback addresses as secure contexts which was previously not present.

All modern browsers (Chrome et. al) support secure cookies on localhost so it only makes sense `tough-cookie` supports this too.


lib/cookie/secureContext.ts
```diff
@@ -0,0 +1,140 @@
+import { URL } from 'url';
+import { isIP, isIPv4, isIPv6 } from 'net';
```
colincasey, line 2: These lines will need to be removed because `tough-cookie` is sometimes used in JavaScript runtimes other than Node.js we don't allow builtin modules to be imported. You can see this in the [CI run for this PR](https://github.com/salesforce/tough-cookie/actions/runs/13434219778/job/37563458160?pr=498) or if you run `npm run lint` locally. 

For the `URL` object, it doesn't need an import as it's already [available as a global](https://developer.mozilla.org/en-US/docs/Web/API/URL) in most JavaScript runtimes.

I'll outline some alternatives you can use instead of `isIP`, `isIPv4`, and `isIPv6` in the relevant areas below.

lib/cookie/cookieJar.ts
```diff
@@ -858,9 +859,7 @@ export class CookieJar {
     const host = canonicalDomain(context.hostname)
     const path = context.pathname || '/'
 
-    const secure =
-      context.protocol &&
-      (context.protocol == 'https:' || context.protocol == 'wss:')
+    const secure = isPotentiallyTrustworthy(url);
```
colincasey, line 862: Since you're already editing this line, could you also rename `const secure` to `const secureConnection` to align with the suggested rename of the `isPotentiallyTrustworthy` function?

lib/cookie/secureContext.ts
```diff
@@ -0,0 +1,140 @@
...
+
+/**
+ * Checks if the given IPv4 address is in the 127.0.0.0/8 loopback range.
+ *
+ * @remarks
+ * @param address - The IPv4 address to check.
+ * @returns `true` if the address is loopback, otherwise `false`.
+ */
```
colincasey, line 10:
```suggestion
```

Documentation is only enforced for functions exposed in the public API. This code is simple enough to be self-describing so the doc comment seems unnecessary.

lib/cookie/secureContext.ts
```diff
@@ -0,0 +1,140 @@
...
+  );
+}
+
+/**
+ * Checks if the given IPv6 address is the loopback address (`::1`).
+ *
+ * @remarks
+ * @param address - The IPv6 address to check.
+ * @returns `true` if the address is loopback, otherwise `false`.
+ */
```
colincasey, line 27:
```suggestion
```

Documentation is only enforced for functions exposed in the public API. This code is simple enough to be self-describing so the doc comment seems unnecessary.

lib/cookie/secureContext.ts
```diff
@@ -0,0 +1,140 @@
...
+  if (isIPv4(address)) {
+    return isLoopbackV4(address);
+  }
+
+  if (isIPv6(address)) {
+    return isLoopbackV6(address);
+  }
+
+  return false;
+}
```
colincasey, line 53: If you accept my suggestion on how to change how IP testing is done, this function will not be required.

lib/cookie/secureContext.ts
```diff
@@ -0,0 +1,140 @@
...
+
+/**
+ * Checks if the given lowercase host ends with the `.localhost` TLD.
+ *
+ * @remarks
+ * @param lowerHost - The lowercase host string to check.
+ * @returns `true` if the host ends with `.localhost`, otherwise `false`.
+ */
```
colincasey, line 61:
```suggestion
```

Documentation is only enforced for functions exposed in the public API. This code is simple enough to be self-describing so the doc comment seems unnecessary.

lib/cookie/secureContext.ts
```diff
@@ -0,0 +1,140 @@
...
+  return lowerHost.endsWith('.localhost');
+}
+
+/**
+ * Checks if the given host is `localhost` or matches `*.localhost`.
+ *
+ * @remarks
+ * @param host - The host string to check.
+ * @returns `true` if the host is considered local, otherwise `false`.
+ */
```
colincasey, line 72:
```suggestion
```

Documentation is only enforced for functions exposed in the public API. This code is simple enough to be self-describing so the doc comment seems unnecessary.

lib/cookie/secureContext.ts
```diff
@@ -0,0 +1,140 @@
...
+    isNormalizedLocalhostTLD(lowerHost);
+}
+
+/**
+ * Removes leading and trailing square brackets if present.
+ * Adapted from https://github.com/chromium/chromium/blob/main/url/gurl.cc#L440-L448
+ *
+ * @param host - The host string to process.
+ * @returns The host string without leading and trailing square brackets if previously present.
+ */
```
colincasey, line 85:
```suggestion
```

Documentation is only enforced for functions exposed in the public API. This code is simple enough to be self-describing so the doc comment seems unnecessary.

lib/cookie/secureContext.ts
```diff
@@ -0,0 +1,140 @@
...
+function hostNoBrackets(host: string): string {
+  if (host.length >= 2 && host.startsWith('[') && host.endsWith(']')) {
```
colincasey, line 87:
```suggestion
  // Adapted from https://github.com/chromium/chromium/blob/main/url/gurl.cc#L440-L448
  if (host.length >= 2 && host.startsWith('[') && host.endsWith(']')) {
```

Inline where you sourced your adaptation.

lib/cookie/secureContext.ts
```diff
@@ -0,0 +1,140 @@
...
+ * A URL is considered potentially trustworthy if it:
+ * - Uses HTTPS or WSS schemes
+ * - Points to a loopback address (IPv4 127.0.0.0/8 or IPv6 ::1)
+ * - Uses localhost or *.localhost hostnames
+ * 
+ * @param inputUrl - The URL string or URL object to check.
+ * @returns `true` if the URL is potentially trustworthy, otherwise `false`.
+ * @see {@link https://w3c.github.io/webappsec-secure-contexts/#potentially-trustworthy-origin W3C Spec}
+ */
+export function isPotentiallyTrustworthy(inputUrl: string | URL): boolean {
```
colincasey, line 105:
```suggestion
export function isSecureConnection(inputUrl: string | URL): boolean {
```

[RFC6265bis-19](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-19) uses the terminology `"secure" connection` when referring to the behaviour being outlined in this PR. E.g.;

> If the request-uri does not denote a "secure" connection (as defined by the user agent), and the cookie's secure-only-flag is true, then abort these steps and ignore the cookie entirely.
> - https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-19#section-5.7-3.13.1

> NOTE: The notion of a "secure" connection is not defined by this document. Typically, user agents consider a connection secure if the connection makes use of transport-layer security, such as SSL or TLS, or if the host is trusted. For example, most user agents consider "https" to be a scheme that denotes a secure protocol and "localhost" to be trusted host.
> - https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-19#section-5.8.3-2.1.2.3.2

This function should reflect that terminology and be renamed to `isSecureConnection`.

lib/cookie/secureContext.ts
```diff
@@ -0,0 +1,140 @@
...
+  let url: URL;
+
+  // try ... catch doubles as an opaque origin check
+  if (typeof inputUrl === 'string') {
+    try {
+      url = new URL(inputUrl);
+    } catch {
+      return false;
```
colincasey, line 113: This line needs a unit test to be added to exercise this `catch` branch.

lib/cookie/secureContext.ts
```diff
@@ -0,0 +1,140 @@
...
+    }
+  } else {
+    url = inputUrl;
```
colincasey, line 116: The accepted `URL` argument type is not represented by any unit test. One needs to be added to validate this branch.

lib/cookie/secureContext.ts
```diff
@@ -0,0 +1,140 @@
...
+    scheme === 'https' ||
+    scheme === 'wss' // https://w3c.github.io/webappsec-secure-contexts/#potentially-trustworthy-origin
+  ) {
+    return true;
+  }
+
+  // If it's already an IP literal, check if it's a loopback address
+  if (isIP(hostname)) {
+    return isIpLoopback(hostname);
+  }
```
colincasey, line 134: To remove the dependency on the `isIP`, `isIPv4`, and `isIPv6`, you could do the following:

```suggestion
  if (IP_V6_REGEX_OBJECT.test(hostname)) {
    return isLoopbackV6(hostname)
  }
  
  if (IP_V4_REGEX_OBJECT.test(hostname)) {
    return isLoopbackV4(hostname)
  }
```


- We already handle IPv6 testing in other areas using the [IP_V6_REGEX_OBJECT](https://github.com/salesforce/tough-cookie/blob/master/lib/cookie/constants.ts#L36)
- We don't have an `IP_V4_REGEX_OBJECT` defined in [constants.ts](https://github.com/salesforce/tough-cookie/blob/master/lib/cookie/constants.ts) but one could be added for this purpose. It should match the implementation used by [`isIPv4` from Node.js](https://github.com/nodejs/node/blob/main/lib/internal/net.js#L18)

lib/__tests__/cookieJar.spec.ts
```diff
@@ -1026,6 +1026,72 @@ describe('CookieJar', () => {
       ])
     })
   })
+
+  describe('Potentially Trustworthy Contexts', () => {
```
colincasey, line 1030: This is missing a test for the `*.localhost` case. It wouldn't hurt to add cases for the `https` and `wss` protocols as well for completeness (even if those are already exercised in the existing test suite).

colincasey (changes requested):

lib/cookie/secureContext.ts
Chriss4123: [Secure Contexts](https://www.w3.org/TR/secure-contexts/) uses the terminology 'potentially trustworthy'. What you described is an actual secure connection, i.e. if the web server has an SSL certificate installed.

When a browser (i.e. Chromium) decides to send `Secure` cookies, it checks the scheme for `https`, `wss`, etc. If it then connects to the server and it doesn't have an SSL certificate the request will fail anyways thus the cookies won't get sent.

`localhost` is an exception to this as it is considered potentially trustworthy.

A potentially trustworthy origin indicates that the scheme is of secure nature and/or we are on localhost. An actual "secure" connection means the web server has an SSL certificate which we can't check.

Therefore I believe `isPotentiallyTrustworthy` is more suitable here as it would be misleading to label it as secure when we in fact don't know that.

lib/cookie/cookieJar.ts
Chriss4123: `secureConnection` would be misleading here as it can be interpreted as having established a connection and done a valid tls/ssl handshake when in fact we are only checking the scheme and/or localhost.

I think `secure` is fine, however I could also do `potentiallyTrustworthy` instead.

awaterma: Could we please consider `potentiallyTrustworthy` or maybe even, `potentiallyTrusted` here? I think `secure` has the same issue as `secureConnection`, that you've identified.

Chriss4123: > Could we please consider `potentiallyTrustworthy` or maybe even, `potentiallyTrusted` here? I think `secure` has the same issue as `secureConnection`, that you've identified.

I like this solution. I'll also add a link to https://www.w3.org/TR/secure-contexts/#is-origin-trustworthy above the line.

lib/cookie/secureContext.ts
colincasey: Understood @Chriss4123, though my intention here was to ground the naming in language referenced in [RFC6265bis-19](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-19) which is deliberately vague about how it's defining this concept. Again, I'll point to the following:

> NOTE: The notion of a "secure" connection is not defined by this document. Typically, user agents consider a connection secure if the connection makes use of transport-layer security, such as SSL or TLS, or if the host is trusted. For example, most user agents consider "https" to be a scheme that denotes a secure protocol and "localhost" to be trusted host.
> - https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-19#section-5.8.3-2.1.2.3.2

Based on the examples given in the above text, the concept of a "secure connection" from this RFC can cover both the scheme and host. 

You point out in your reply, that this is inaccurate and the terminology from [Secure Contexts](https://www.w3.org/TR/secure-contexts/) is more suitable. I don't disagree with you but I still prefer the terminology of "secure connection" to be used in the code because:
- In [RFC6265bis-19](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-19) you will find the phrase `"secure" connection` referenced at multiple points in the implementation details in both the Retrieval Algorithm and the Storage Model where the code from your PR is called
- The phrase `potentially trustworthy origin` can be found buried in [Appendix A of RFC6265bis-19](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-19#appendix-A-1.7.1) but there's no mention of [Secure Contexts](https://www.w3.org/TR/secure-contexts/) in the list of references. 

So I would ask you to consider this from the perspective of a maintainer. It's easier to compare the implementation to the RFC when the language used in both matches. Even if that language is vague and somewhat misleading (which the RFC explicitly calls out) 😅  

Mind you, I'm not going to die on this hill. There's enough overlap in the terminology to backup both arguments (https://github.com/httpwg/http-extensions/issues/2605, [httpwg/http-extensions#2759](http://github.com/httpwg/http-extensions/pull/2759)). If we keep the naming of `isPotentiallyTrustworthy` then some extra comments will need to be added to connect the dots between how [Secure Contexts](https://www.w3.org/TR/secure-contexts/) relates to the implementation details of [RFC6265bis-19](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-19).

lib/cookie/cookieJar.ts
colincasey: Yes, I had considered maybe leaving it as just `secure` is fine but I think `secureConnection` would help to disambiguate between the cookie's `Secure` attribute and what the RFC is referring to here (`"secure" connection`) in the Storage Model implementation details. 

Let's put a pin on this for now though while we discuss the overall terminology in https://github.com/salesforce/tough-cookie/pull/498#discussion_r1964293125

lib/cookie/secureContext.ts
Chriss4123: > Understood @Chriss4123, though my intention here was to ground the naming in language referenced in [RFC6265bis-19](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-19) which is deliberately vague about how it's defining this concept. Again, I'll point to the following:
> 
> > NOTE: The notion of a "secure" connection is not defined by this document. Typically, user agents consider a connection secure if the connection makes use of transport-layer security, such as SSL or TLS, or if the host is trusted. For example, most user agents consider "https" to be a scheme that denotes a secure protocol and "localhost" to be trusted host.
> > 
> > * https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-19#section-5.8.3-2.1.2.3.2
> 
> Based on the examples given in the above text, the concept of a "secure connection" from this RFC can cover both the scheme and host.
> 
> You point out in your reply, that this is inaccurate and the terminology from [Secure Contexts](https://www.w3.org/TR/secure-contexts/) is more suitable. I don't disagree with you but I still prefer the terminology of "secure connection" to be used in the code because:
> 
>     * In [RFC6265bis-19](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-19) you will find the phrase `"secure" connection` referenced at multiple points in the implementation details in both the Retrieval Algorithm and the Storage Model where the code from your PR is called
> 
>     * The phrase `potentially trustworthy origin` can be found buried in [Appendix A of RFC6265bis-19](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-19#appendix-A-1.7.1) but there's no mention of [Secure Contexts](https://www.w3.org/TR/secure-contexts/) in the list of references.
> 
> 
> So I would ask you to consider this from the perspective of a maintainer. It's easier to compare the implementation to the RFC when the language used in both matches. Even if that language is vague and somewhat misleading (which the RFC explicitly calls out) 😅
> 
> Mind you, I'm not going to die on this hill. There's enough overlap in the terminology to backup both arguments ([httpwg/http-extensions#2605](https://github.com/httpwg/http-extensions/issues/2605), [httpwg/http-extensions#2759](https://github.com/httpwg/http-extensions/pull/2759)). If we keep the naming of `isPotentiallyTrustworthy` then some extra comments will need to be added to connect the dots between how [Secure Contexts](https://www.w3.org/TR/secure-contexts/) relates to the implementation details of [RFC6265bis-19](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-19).

 RFC6265bis-19 deliberately avoids rigidly defining what constitutes a “secure” connection for cookies, instead deferring that decision to user agents. In the spec’s [definition](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis#section-4.1.2.5) of the `Secure` attribute, it explicitly notes that **“’secure’ is defined by the user agent”**. The cookie [retrieval](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis#section-5.8.3) algorithm also requires that a `Secure` cookie be sent only if the request’s URL is over a “secure” channel – again **left to the UA’s definition**. An informative note in the spec:

> *“NOTE: The notion of a ‘secure’ connection is not defined by this document. Typically, user agents consider a connection secure if the connection makes use of transport-layer security, such as SSL or TLS, or if the host is trusted. For example, most user agents consider ‘https’ to be a scheme that denotes a secure protocol and ‘localhost’ to be [a] trusted host.”* ([draft-ietf-httpbis-rfc6265bis-19](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis#name-retrieval-algorithm))

This makes it clear the IETF spec intentionally refrains from hard-coding a definition of “secure connection.” Instead, it acknowledges common UA behavior (treating TLS-protected schemes as secure, and even treating certain **trusted hosts like `localhost` as secure-equivalent**), but leaves the exact criteria up to implementations.

RFC6265bis-19 expects browsers to apply their own notion of what is secure. Modern browsers have, in practice, converged on the concept of **“potentially trustworthy origins”** to make this determination, rather than a naive check of scheme or transport alone.

## Potentially Trustworthy Origins
Broadly, the phrase *“secure connection”* is less precisely defined, whereas **“potentially trustworthy origin”** (and the related concept of **“secure contexts”**) is the well-defined, normative standard used to gate security-sensitive behaviors. Multiple authoritative specifications define and use the concept of “potentially trustworthy” in place of ad-hoc notions of security:

- **W3C Secure Contexts Specification:** This spec defines what it means for an origin to be *potentially trustworthy*. It provides an [algorithm](https://www.w3.org/TR/secure-contexts/#is-origin-trustworthy) enumerating conditions under which an origin is considered secure enough. The Secure Contexts spec’s [core definition](https://www.w3.org/TR/secure-contexts/#is-origin-trustworthy) is: *“A potentially trustworthy origin is one which a user agent can generally trust as delivering data securely.”* This formalizes the intuitive idea mentioned in RFC6265bis’s note [that `https` and `localhost` are considered secure](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis#name-retrieval-algorithm) into a normative algorithm.

- **WHATWG HTML Standard:** HTML integrates the secure contexts concept to determine when a browsing context is a “secure context.” The HTML spec states that *“an environment is a secure context if the following algorithm returns true…”*, and one of the key steps in that [algorithm](https://html.spec.whatwg.org/multipage/webappapis.html#secure-contexts) is: *“If the result of **Is URL potentially trustworthy?** given *environment’s top-level creation URL* is ‘Potentially Trustworthy’, then return true.”*. This defers to the Secure Contexts spec’s [“Is origin potentially trustworthy?”](https://w3c.github.io/webappsec-secure-contexts/#potentially-trustworthy-url) check for the top-level page’s origin to decide if a context is secure.

- **WHATWG Fetch Standard:** The Fetch spec similarly uses “potentially trustworthy” in defining network request behaviors. For example, the algorithms for handling Mixed Content use this terminology: *“Upgrade a mixed content request to a potentially trustworthy URL, if appropriate.”* ([Fetch Standard - WhatWG](https://fetch.spec.whatwg.org/#main-fetch)).

Summarized: **Secure Contexts, HTML, and Fetch (along with related specs like Mixed Content) all treat “potentially trustworthy origin” as the canonical definition of a secure environment.** This covers HTTPS and WSS, but also explicitly includes other cases like `localhost`, and excludes others (plain `http` to remote hosts) – thus providing a precise security model. The term *“secure connection”* by itself has no formal definition in these specs and could be interpreted in narrower ways (e.g. “only TLS”).

## Browser Implementations
Browser behavior has evolved to align with the **“potentially trustworthy origin”** model as it’s more accurate and future-proof than a simplistic “secure connection” check. Both Chromium and Firefox have made conscious changes to use the broader *trustworthy origin* concept for cookies marked Secure, rather than just checking for HTTPS:

- **Firefox:** Firefox historically allowed Secure cookies on `https://` but not on `http://` origins. However, as developers increasingly run local servers on `http://localhost` and expect to use Secure cookies (or the `__Host-`/`__Secure-` name prefixes which require secure contexts), Mozilla updated Firefox’s implementation to treat **`http://localhost` as secure for cookies**. The Firefox bug report states the plan: *“By RFC6265, the definition of ’secure’ [protocol] is defined by the user-agent. I propose to unify the implementations and use `nsMixedContentBlocker::IsPotentiallyTrustworthyOrigin()` instead of a simple ’https’ scheme check.”* ([1618113 - Allow ’secure’ cookies when set by localhost](https://bugzilla.mozilla.org/show_bug.cgi?id=1618113)). The code for this can be found [here](https://github.com/mozilla/gecko-dev/blob/af5960c7041d895174ea6bceb1822773e0c9b350/dom/security/nsMixedContentBlocker.cpp#L298-#L366).

- **Chromium:** Historically, Chrome **did** allow `Secure` cookies on `http://localhost` (treating localhost as secure), but it initially did **not** allow the stricter `__Host-` and `__Secure-` cookie name prefixes on localhost. The Chromium team has broadened the definition of “secure” for cookies to include *potentially trustworthy origins* as well. A Chromium [issue discussion](https://issues.chromium.org/issues/40196122) notes that they consider support for `__Host-`/`__Secure-` on localhost **“in scope”**. The cookie handling code defines an enumeration of cookie source schemes with categories for **cryptographic vs. non-cryptographic but trustworthy**. Chrome’s [`CookieAccessScheme`](https://chromium.googlesource.com/chromium/src/+/HEAD/net/cookies/cookie_constants.h#L265) includes: `kCryptographic` (for HTTPS/WSS, implicitly trustworthy) and `kTrustworthy` (for non-HTTPS schemes that are deemed potentially trustworthy). A [comment](https://chromium.googlesource.com/chromium/src/+/HEAD/net/cookies/cookie_constants.h#L264) in the source links directly to the W3C Secure Contexts definition of *potentially trustworthy origin*, showing that Chrome’s engineers explicitly built the **Secure=“potentially trustworthy”** logic into their implementation...

It’s clear that “secure connection” as a term is too simplistic to capture the real conditions under which cookies (and other features) should be considered secure. A strict reading might equate *secure connection* only with **TLS** however both Firefox and Chrome purposefully include **“localhost” (no TLS) as secure** based on the reasoning that it’s a trustworthy origin. The term *potentially trustworthy origin* **precisely describes the broader notion of security** – it includes *both* cryptographic transport *and* other trustworthy circumstances (like loopback interfaces, local files, or extension contexts).

## Terminology and Maintainability
The phrase “secure connection” is ambiguous without an explicit definition. The [RFC6265bis](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis#section-5.8.3-2.1.2.3.2) text itself had to add a clarifying note to hint at meaning. If our implementation simply says “if secure connection, then …” we’d likely need a similar comment explaining what “secure” entails. By using the term *“potentially trustworthy origin”* (or referring to a *secure context*), we immediately invoke a well-defined concept. It’s unambiguous to a web platform engineer that this means the origin satisfies the standard [Secure Contexts](https://www.w3.org/TR/secure-contexts/#is-origin-trustworthy) criteria. I believe future contributors are more likely to be familiar with the Secure Contexts definition (since it’s used across many specs and APIs) than with the cookie spec’s isolated wording.

Most modern web platform code (both in specs and in browser source) has standardized on talking about “trustworthy” or “secure contexts.” Our cookie implementation logic should speak the same language, aligning itself conceptually with web platforms, and making the code self-documenting in a clearer way.

It’s important to note that using *“potentially trustworthy origin”* in our implementation detail is **not actually contradicting RFC6265bis-19** – it’s clarifying it. The RFC’s intent was clearly that user agents decide what “secure” means. Today’s consensus is that *secure = potentially trustworthy origin*. [Issue discussions](https://github.com/httpwg/http-extensions/issues/2605) in the IETF HTTP Working Group suggest exactly this alignment: to *“consider reusing the notion of ‘Is origin potentially trustworthy?’ from the Secure Contexts spec, instead of [a] user-agent-dependent ‘secure protocol’ check”*. So while our implementation detail might use a different phrasing than the raw RFC text, it actually **reflects the RFC’s prescribed approach (deferring to UA judgment) more faithfully** by naming the precise criterion we’ve chosen to follow.

Sticking slavishly to the term “secure connection” for the sake of mirroring the RFC introduces ambiguity, require redundant explanatory comments, and risk drifting from the rest of the web platform’s security model.

## Conclusion 
**Using “potentially trustworthy” in place of “secure connection”** in our RFC6265bis-19 cookie jar implementation is a technically sound decision that is inline with the reality of web security today. The RFC deliberately empowers user agents to decide what “secure” means; adopting the well-defined *potentially trustworthy origin* standard is the best interpretation of that mandate.

I can make appropriate comments linking the two terms thus maintaining fidelity to RFC6265bis-19 and simultaneously ensure our implementation is clear, consistent, and maintainable for the future. Adopting *“potentially trustworthy origin”* terminology will strengthen both the technical correctness and the clarity of our cookie security model, which is an unequivocal win for people reading the code and implementers.

Chriss4123:
@colincasey I've completed all the requested changes and also added a comment linking a potentially trustworthy origin and the notion of a "secure" connection in [draft-ietf-httpbis-rfc6265bis-19](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-19#section-5.8.3-2.1.2.3).

I'm ready for you to re-review the PR.

lib/cookie/secureContext.ts
Chriss4123: Sorry about the essay, I honestly just looked at all my research I did before making this change and wrote about all of it and when I realized it was too long it was too late as I was going out that night so I though "oh well, it is what it is" :P

Chriss4123:
Will look at the integration tests later and find what is causing these issues. Will let you know once all is good.

colincasey:
Thanks @Chriss4123, I'll hold off on approval until it's clearer why `jsdom` tests would fail with this code included.

Chriss4123:
@colincasey The reason `jsdom` tests fail is because they expect `Secure` cookies to **not** work on localhost, as seen [here](https://github.com/jsdom/jsdom/blob/8b62a4f474664ccd6d84ea63f1697a6bd90e638a/test/api/cookies.js#L261-L287) and in other cookie tests.

The `testHost + "/TestPath/set-cookie-from-server"` path sets multiple cookies, one of which being a cookie with the `Secure` attribute (`testHost` is `http://127.0.0.1:...`) and the assertion expects the `Secure` cookie to not be present:
```js
assertCookies(cookieHeader, [
  "Test1=Basic",
  "Test2=PathMatch",
  "Test6=HttpOnly",
  // "Test7=Secure; expires=Wed, 13-Jan-2051 22:23:01 GMT; path=/; Secure", (This cookie is present)
  "Test9=Duplicate",
  "Test10={\"prop1\":5,\"prop2\":\"value\"}",
  "Malformed"
]);
```

due to the scheme being `http` instead of `https`.

This is unfortunately something out of our control, but as their goal states:
> the goal of the project is to emulate enough of a subset of a web browser to be useful for testing and scraping real-world web applications.

I do believe that part of that should be accepting `Secure` cookies on `localhost` as many browsers (Firefox, Chromium, etc) do.

I'll open a PR in `jsdom` attempting to change that but in case users don't want this functionality we can add a new parameter in the get cookies function that explicitly controls whether we should treat localhost/loopback addresses as trustworthy, something like `allowSecureOnLocal`.

As it stands `jsdom` isn't automatically using the latest version of `tough-cookie`, thus we can add this parameter, merge the PR and raise attention to the `jsdom` devs that if they'd like to keep old cookie trustworthy functionality they need to set `allowSecureOnLocal` to `true`.

What do you think of this?

Chriss4123:
By the way, when I run `npm run prepare-pr` it replaces all `CRLF`'s with `LF`'s in the docs:
> In Unix systems the end of a line is represented with a line feed (LF). In windows a line is represented with a carriage return (CR) and a line feed (LF) thus (CRLF). when you get code from git that was uploaded from a unix system they will only have an LF.

(taken from https://stackoverflow.com/a/5834094)

This is because I am on Linux and the last person to run `prepare-pr` likely did it on Windows. I am going to commit it, but if this is an issue we can revert this commit and make a plan to use `CRLF` instead.

colincasey:
> By the way, when I run npm run prepare-pr it replaces all CRLF's with LF's in the docs

This is expected @Chriss4123. There's a bug in the [api-extractor](https://github.com/microsoft/rushstack/issues/1467) that seems to be at the root of this. Our automation doc checks should catch if there's any problems here so if that passes you should be all good.

As for the code, I have one question for you on the addition of `allowSecureOnLocal` - do you think that option should be  available on the [`CookieJar` constructor](https://github.com/salesforce/tough-cookie/blob/master/api/docs/tough-cookie.createcookiejaroptions.md) instead of as an option on the `getCookies` method? The RFC states this should be at the discretion of the user agent during both [storage](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-19#section-5.7-3.13.1) and [retrieval](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-19#section-5.8.3-2.1.2.3.1) so I find it a bit odd that it's only exposed as configuration when retrieving cookies.

Chriss4123:
That is a great point @colincasey. The only reason I didn't do this is because the cookie jar _doesn't_ actually follow this step in the algorithm:
```js
> jar.setCookie('ipv6cookie=ipv6val; Secure; Path=/', 'http://[::ffff:7f00:1]:5000/')
Promise {
  Cookie="ipv6cookie=ipv6val; Path=/; Secure; hostOnly=true; aAge=0ms; cAge=0ms",
  [Symbol(async_id_symbol)]: 137,
  [Symbol(trigger_async_id_symbol)]: 6
}

> jar.store.idx
[Object: null prototype] {
  '::ffff:7f00:1': [Object: null prototype] {
    '/': [Object: null prototype] {
      ipv6cookie: Cookie="ipv6cookie=ipv6val; Path=/; Secure; hostOnly=true; aAge=142897ms; cAge=142897ms"
    }
  }
}
```
If it did, this cookie wouldn't be present in the store at all. Testing with this code:
```py
from flask import Flask, make_response, request

app = Flask(__name__)

@app.route('/setcookie')
def set_cookie():
    resp = make_response('')
    resp.set_cookie('example', '', secure=True)
    return resp

@app.route('/getcookie')
def get_cookie():
    return 'Cookie is set' if 'example' in request.cookies else 'No cookie set'

if __name__ == '__main__':
    app.run()
```

and Firefox on `[::ffff:7f00:1]:5000`, I can see that the `example` cookie is **not** set when I visit `/setcookie`, however this library **will** set, albeit it wont be returned so in the end it's **as if it was never set** if you only use the `getCookies` functions and don't access the store directly.

If you think this is a flaw too, let me know and I can open a new PR for it.
I'll also adjust the logic within this PR to reflect this rule in the algorithm by moving the `allowSecureOnLocal` to the constructor of `CookieJar`.

colincasey:
> this library will set, albeit it wont be returned so in the end it's as if it was never set

This is what I was hinting at. I noticed in the tests added to support the `allowSecureOnLocal ` option you were able to set the cookie in the store but not retrieve it. While I don't think that's technically wrong, I also don't think it's technically right 😅 

What do you think of the following alternatives?

#### Expose the `allowSecureOnLocal` option on the `CookieStore` 

This would allow `setCookie` and `getCookies` to enforce the option and cover what's described in the storage and retrieval algorithms of the RFC. Then, a `CookieJar` with the `allowSecureOnLocal` option disabled would never allow **new** secure `localhost` cookies to be set while also preventing **existing** secure `localhost` cookies that may already be in the store from being returned.

Testing the retrieval case is a bit more complex but can be done if `setCookie` is bypassed and a secure `localhost` cookie is inserted directly into the store.

#### Expose the `allowSecureOnLocal` option on `setCookie` and `getCookies`

Same as the above, this would allow `setCookie` and `getCookies` to enforce the option and cover what's described in the storage and retrieval algorithms of the RFC. This finer degree of control comes with the trade-off of increased configuration complexity for the API with little benefit. Other then testing, I can't imagine a scenario where you'd want to use different values for `allowSecureOnLocal` when storing vs. retrieving.

Chriss4123:
I prefer having the option in the constructor because while I like granular control I don't see a scenario where you'd want localhost as both trustworthy and not trustworthy. You'd also have to specifically say that localhost isn't trustworthy each time you set/get a cookie which would be a nuisance.

What's your thoughts on the issue I pointed out regarding setting `Secure` cookies on a non trustworthy address?

It is definitely more technically correct to **abort** as defined in the algorithm but the final call is obviously up to you.

Chriss4123:
> This would allow `setCookie` and `getCookies` to enforce the option and cover what's described in the storage and retrieval algorithms of the RFC. Then, a `CookieJar` with the `allowSecureOnLocal` option disabled would never allow **new** secure `localhost` cookies to be set while also preventing **existing** secure `localhost` cookies that may already be in the store from being returned.

@colincasey Sorry, didn't realize you actually answered my question, guess I was too tired 😅.
Anyways, I've gone ahead and implemented that parameter and also the logic to abort if we're setting a secure cookie on a non-secure URL.

In the process, I had to modify multiple current and legacy tests to reflect that setting a cookie with the `Secure` attribute on a non-secure URL now throws an error instead of succeeding.

Previously, those tests expected no error when setting the cookie and instead not retrieving the cookie.

colincasey:
@Chriss4123 discussed this PR with the other maintainers today. We're looking at shipping this in a v6 release. @awaterma is also going to give it a review and we're just figuring out if there are other work items that we'd like to include in a new major release so don't be surprised if it takes a while longer for it to be merged. Thanks so much for this contribution!

Chriss4123:
@colincasey got it. For now I can just use my branch in my projects where I need this functionality.

If you need any further changes let me know.

Do you have a broad ETA in mind for this being merged (i.e. v6 release)?

colincasey (dismissed):

colincasey:
@Chriss4123 the ETA would definitely be a broad one 😅  My guess is somewhere between 2 - 4 weeks.

Chriss4123:
@colincasey that ETA is definitely quicker than I anticipated which is good to hear!

I realized I made a small mistake on a test case:

Pre `Secure` cookie fail when not a potentially trustworthy origin:
```ts
it('should fail silently when no Domain or Path', async () => {
  await cookieJar.setCookie('__Host-SID=12345; Secure', insecureUrl, {})
  const cookies = await cookieJar.getCookies(insecureUrl)
  expect(cookies).toEqual([])
})
```
Post `Secure` cookie fail when not a potentially trustworthy origin:
```ts
it('should fail silently when no Domain or Path', async () => {
  await cookieJar.setCookie('__Host-SID=12345', insecureUrl, {})
  const cookies = await cookieJar.getCookies(insecureUrl)
  expect(cookies).toEqual([])
})
```

I think there was an issue with the initial test case: `__Host-` prefix requires the `Secure` attribute and thus a secure URL so I'm not 100% sure why `insecureUrl` was being used, which led to me just getting rid of the `Secure` attribute altogether as an error is now raised when a `Secure` cookie is paired with an unsecure URL. I've updated it to use the `Secure` attribute with `secureUrl` (so it actually tests that it silently fails when no Domain or Path is set).

colincasey (approved):

colincasey:
@Chriss4123 we're going to include this in a new major release (`v6`). I'm going to publish a pre-release sometime next week to give consumers a chance to try things out before it becomes official. Thanks so much for this contribution!

Chriss4123:
> @Chriss4123 we're going to include this in a new major release (`v6`). I'm going to publish a pre-release sometime next week to give consumers a chance to try things out before it becomes official. Thanks so much for this contribution!

Happy this has been merged! On the `jsdom` side do you think I should open an issue as a precaution to disable the allow secure on localhost flag to keep the old behavior or do you think it's unnecessary as they will likely figure it out if they try bump `tough-cookie`s version?

Apart from that happy I could help!

colincasey:
@Chriss4123 Once I publish a pre-release, it'll be possible to open PRs against `jsdom` with `tough-cookie` set to that new version with options to either:
- Update their test suite to use secure domains in their tests or allow the localhost cookies
- Configure their internal `CookieJar` to use the opt-out flag `allowSecureOnLocal = false` to retain their existing test behavior

I think that would be useful in getting a conversation started around how this change affects `jsdom` and what they'd like to see done to support this change. 

I don't mind doing that work but I also won't stop you if you wanted to tackle that before I get around to it 😄

ertl:
@colincasey When is the official release of version 6.x scheduled? Are there any known downsides to using the release candidate? We're considering it urgently due to the need for the new secure behavior.

colincasey:
@ertl There are no major downsides to using the release candidate. 

There is one minor detail that I've got to sort out first though and it's whether to enforce the secure behavior on both cookie storage and retrieval. [RFC 6265](https://datatracker.ietf.org/doc/html/rfc6265#section-5.3) doesn't state it should be enforced during storage but [httpbis-rfc6265bis (draft 20)](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-20#section-5.7-3.13.1) does and that is currently causing an issue with [one of the web platform tests for websockets](https://wpt.live/websockets/cookies/006.html) that [jsdom uses](https://github.com/jsdom/jsdom/blob/main/Contributing.md#web-platform-feature-tests). 

Removing the [code here](https://github.com/salesforce/tough-cookie/blob/master/lib/cookie/cookieJar.ts#L611-L626) is more in alignment with RFC 6265 and this also seems to satisfy the web platform tests. I just haven't had the time yet to double-check the expected websocket cookie behavior in different browsers to confirm if that's the correct action to take or not.

ertl:
@colincasey Thanks for the update. We'll go ahead and give the release candidate a try.

PR #506, opened 2025-03-07
use ESM instead of CJS

wjhsf:
This PR changes our distributed code from CJS to ESM! To do so, we had to remove the `vows` tests and add `.js` file extensions to all of our imports. Other than that, just a few minor adjustments to `eslint.config.mjs`, `jest.config.js`, and `package.json`.

(A new PR instead of the older #435 so that I don't have to deal with merge conflicts.)


package.json
```diff
@@ -96,7 +97,7 @@
     "lint": "npm run _lint:check",
     "prepack": "npm run build",
     "prepare-pr": "npm test && npm run _api:update && npm run _docs:generate && npm run _format:fix && npm run _lint:fix",
```
colincasey, line 99:
```suggestion
    "prepare-pr": "npm run build && npm test && npm run _api:update && npm run _docs:generate && npm run _format:fix && npm run _lint:fix",
```

Since `npm run build` has been removed from the `test` script it needs to be added to the `prepare-pr` script. The doc tasks won't work right if there is no `dist` directory.

colincasey (dismissed):
So nice to see the legacy test suite go away 😄

colincasey (approved):

awaterma (approved):
LGTM to me as well.

PR #510, opened 2025-04-04
Switch from jest to vitest

wjhsf:
Vitest has better ESM support than jest and is faster. Running tests 10x on my local machine took 24 seconds with jest, but only 9 seconds with vitest.

Since vitest requires ESM, this PR is built on #506.


lib/__tests__/cookieToAndFromJson.spec.ts
```diff
@@ -1,15 +1,20 @@
+import { describe, expect, it } from 'vitest'
 import { Cookie } from '../cookie/cookie.js'
```
colincasey, line 3: The Jest fake timers were [enabled globally](https://github.com/salesforce/tough-cookie/blob/master/jest.config.ts#L9) since we often need finer control over date-based testing. It might be worthwhile to do something similar for `vitest` but, barring that, adding the following will help to resolve the need for capturing and comparing a `[start, end]` date for these tests which did fail several times for me locally.

```suggestion
import { beforeEach, describe, expect, it, vi } from "vitest";
import { Cookie } from "../cookie/cookie.js";

beforeEach(() => {
  vi.useFakeTimers()
})
```

lib/__tests__/cookieToAndFromJson.spec.ts
```diff
@@ -1,15 +1,20 @@
...
     )
+    const finish = new Date().toISOString()
     if (!cookie) {
       throw new Error('This should not be undefined')
     }
     expect(cookie.toJSON()).toEqual({
-      creation: new Date().toISOString(),
+      // Sometimes we roll over a millisecond, so we need to check both
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+      creation: expect.toBeOneOf([start, finish]),
```
colincasey, line 17: With fake timers this can be:

```suggestion
      creation: new Date().toISOString(),
```

colincasey (changes requested):
A couple small changes here @wjhsf but overall this looks good. Very impressed with the speed compared to Jest 😄

lib/__tests__/cookieUtils.spec.ts
```diff
@@ -1,3 +1,4 @@
+import { describe, expect, it } from 'vitest'
```
colincasey, line 1: This test file looks like it might also need fake timers enabled.

lib/__tests__/jarSerialization.spec.ts
```diff
@@ -173,12 +175,16 @@ describe('cookieJar serialization', () => {
   describe('with a moderately-sized store', () => {
     let jar: CookieJar
     let expires: Date
+    let start: string
```
colincasey, line 178: Let's use fake timers here as well to control creation / lastAccessed timestamps.

package.json
```diff
@@ -96,8 +96,8 @@
     "build": "npm run _build:clean && npm run _build:compile",
     "lint": "npm run _lint:check",
     "prepack": "npm run build",
-    "prepare-pr": "npm run build && npm test && npm run _api:update && npm run _docs:generate && npm run _format:fix && npm run _lint:fix",
-    "test": "jest",
+    "prepare-pr": "npm test && npm run _api:update && npm run _docs:generate && npm run _format:fix && npm run _lint:fix",
```
colincasey, line 99: We still need `npm run build` before the api and doc steps.

colincasey (approved):

colincasey (approved):

PR #536, opened 2025-07-31
Support publishing of both ESM and CJS

colincasey:
The [ESM supported backported to Node 20.19.0](https://github.com/nodejs/node/releases/tag/v20.19.0) is nice but not sufficient for supporting users on earlier versions of Node v20 which is still in active maintenance until 2026-04-30. This PR adds in dual-publishing of both ESM and CJS and ensures that entrypoints and types are correctly declared.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -47,6 +47,20 @@ jobs:
       - name: Run tests
         run: npm test
 
+  esm_cjs:
+    name: Types & Entrypoints
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: lts/*
+          cache: npm
+      - name: Verify types & entrypoints
+        run: npx --yes @arethetypeswrong/cli --pack .
+
   api:
     name: API & Doc Validation
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -61,7 +61,7 @@ jobs:
       - name: Install dependencies
         run: npm ci
       - name: Verify types & entrypoints
-        run: npx --yes @arethetypeswrong/cli --pack .
+        run: npm run _lint:types
 
   api:
     name: API & Doc Validation
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -58,6 +58,8 @@ jobs:
         with:
           node-version: lts/*
           cache: npm
+      - name: Install dependencies
+        run: npm ci
       - name: Verify types & entrypoints
         run: npx --yes @arethetypeswrong/cli --pack .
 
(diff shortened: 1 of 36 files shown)


package.json
```diff
@@ -85,23 +85,31 @@
...
+    "build": "tsup lib/cookie/index.ts --format cjs,esm --dts --clean --sourcemap",
     "lint": "npm run _lint:check",
     "prepack": "npm run build",
     "prepare-pr": "npm run build && npm test -- run && npm run _api:update && npm run _docs:generate && npm run _format:fix && npm run _lint:fix",
     "test": "vitest",
     "version": "npm run _version:generate && npm run prepare-pr && git add --renormalize .",
     "_api:check": "api-extractor run --verbose",
     "_api:update": "api-extractor run --verbose --local",
-    "_build:clean": "rm -rf dist",
     "_build:compile": "tsc",
```
wjhsf, line 113:
```suggestion
```
Also no longer used (looks like `tsup --dts` does type checking).

wjhsf (approved):
Can you add tests that `require('tough-cookie')` and `await import('tough-cookie')` both work?

.github/workflows/ci.yaml
```diff
@@ -47,6 +47,22 @@ jobs:
...
+        uses: actions/checkout@v4
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: lts/*
+          cache: npm
+      - name: Install dependencies
+        run: npm ci
+      - name: Verify types & entrypoints
+        run: npx --yes @arethetypeswrong/cli --pack .
```
wjhsf, line 64: This will always pull latest and run it. We should add it as a regular dependency so that we don't spontaneously combust when they release a breaking change.

wjhsf, line 64: This step is just another test, so it could get shoved into the `test` matrix and we could skip the extra `npm ci`. But then we'd need a conditional skip for jus that step to only run it once, instead of for everything in the matrix. 🤔

wjhsf, line 64: Using `attw` isn't a direct check of `require("tough-cookie")` and `import("tough-cookie")` in node. But I guess neither is vitest, since it has its extra layer on top for mocking. Is this good enough? 🤷

wjhsf (dismissed):

.github/workflows/ci.yaml
colincasey: I suppose I'm going off the fact that `tsup` **should** produce valid ESM and CJS artifacts and `attw` **should** verifies if your types and entrypoints are configured correctly. Maybe it's not a true test since the `require` and `import` are not verified directly but it seems reasonable to trust that these tools are working as advertised.

Setting up a full integration test harness just to verify the `require` and `import` seems a bit heavy. This is why I tried verifying locally then setting up some of the above checks as a safe guard.

How would you prefer we test this @wjhsf?

colincasey: I had considered that but it's clearer what the issue is if this separate job fails + it doesn't need to be checked for everything in the matrix.

wjhsf (approved):

PR #552, opened 2025-12-03
test(timing): retry flaky test and increase threshold

wjhsf:
In #546, we upgraded vitest from v3 to v4. I'm not sure why, but this broke the test that asserts the parsing speed of short vs long strings. Now the ratio is above the threshold about 80% of the time (on my machine), being consistently in the range of 300-400.

This PR increases the threshold slightly (to 300 instead of 250) and retries the one test up to three times if it fails. The tests reliably passed (50/50 runs) with this configuration.


awaterma (approved):
I think we may want to increase this a little bit more, like `350`? But I'm good with this either way.

Issue #582, opened 2026-03-13
Introduce CookiePath nominal type


colincasey:
Part of #581.

## Motivation

RFC 6265 §5.2.4 defines the cookie-path as always starting with `/`. The `defaultPath()` function guarantees this, but `pathMatch()` accepts plain `string` with no enforcement. There is no type-level distinction between a request path, a cookie path, and an arbitrary string.

### Current signatures

```typescript
// Returns a valid cookie path (always starts with "/") but typed as plain string
function defaultPath(path?: Nullable<string>): string

// Both parameters should start with "/" but nothing enforces this
function pathMatch(reqPath: string, cookiePath: string): boolean

// Cookie class — path should start with "/" if non-null
class Cookie {
  path: string | null
}
```

## Nominal type design

```typescript
declare const tag: unique symbol;

export type CookiePath = string & { readonly [tag]: true };

export namespace CookiePath {
  /** Construct a CookiePath from a string, validating the "/" prefix. */
  export function parse(input: string): CookiePath | undefined {
    if (!input.startsWith('/')) return undefined;
    return input as CookiePath;
  }

  /** Compute the default cookie path from a request URI path (RFC 6265 §5.1.4). */
  export function fromRequestPath(path?: Nullable<string>): CookiePath {
    // Existing defaultPath() logic
  }
}
```

### Internal pure algorithm (new)

```typescript
function pathMatchInternal(reqPath: CookiePath, cookiePath: CookiePath): boolean {
  // Pure RFC 6265 §5.1.4 path-match algorithm
}
```

### Public wrapper (unchanged signature)

```typescript
export function pathMatch(reqPath: string, cookiePath: string): boolean {
  // Delegate to internal after parsing
}
```

### Return type narrowing (non-breaking)

| Function | Current return | New return | Breaking? |
|---|---|---|---|
| `defaultPath()` | `string` | `CookiePath` | **No** — assignable to `string` |
| `permutePath()` | `string[]` | `CookiePath[]` | **No** |

### Unchanged public surfaces

- `Cookie.path` remains `string | null`
- `pathMatch()` parameter types remain `string`
- `Store` method `path` parameters remain `Nullable<string>`

PR #598, opened 2026-04-27
feat: introduce CookiePath nominal type

colincasey:
## Summary

- Adds a `CookiePath` branded type and const object with `ROOT`, `parse`, `parentPath`, `defaultPath`, `match`, and `permute`
- Enforces the RFC 6265 §5.1.4 `/`-prefix invariant at compile time via a nominal type pattern
- Purely additive — no changes to existing code

Part 1 of 3 for #582. Stacked on `master`.

Next: #cookie-path-deprecate-wrappers (deprecates existing wrapper functions)

### DIFF
--- a/lib/__tests__/cookiePath.spec.ts
+++ b/lib/__tests__/cookiePath.spec.ts
@@ -1,0 +1,29 @@
+import { describe, expect, it } from 'vitest'
+import { CookiePath } from '../cookie/cookiePath.js'
+
+describe('CookiePath', () => {
+  describe('ROOT', () => {
+    it('equals "/"', () => {
+      expect(CookiePath.ROOT).toBe('/')
+    })
+  })
+
+  describe('parse', () => {
+    it.each([
+      { input: '/', expected: '/' },
+      { input: '/foo', expected: '/foo' },
+      { input: '/foo/bar', expected: '/foo/bar' },
+      { input: '/foo/bar/', expected: '/foo/bar/' },
+    ])('returns CookiePath for valid input "$input"', ({ input, expected }) => {
+      expect(CookiePath.parse(input)).toBe(expected)
+    })
+
+    it.each([
+      { input: '', description: 'empty string' },
+      { input: 'foo', description: 'no leading slash' },
+      { input: 'bar/baz', description: 'relative path' },
+    ])('returns undefined for $description', ({ input }) => {
+      expect(CookiePath.parse(input)).toBeUndefined()
+    })
+  })
+})
--- a/lib/__tests__/cookiePath.spec.ts
+++ b/lib/__tests__/cookiePath.spec.ts
@@ -2,6 +2,7 @@ import { describe, expect, it } from 'vitest'
 import { CookiePath } from '../cookie/cookiePath.js'
 import { defaultPathCases } from './data/defaultPathCases.js'
 import { pathMatchCases } from './data/pathMatchCases.js'
+import { permutePathCases } from './data/permutePathCases.js'
 
 describe('CookiePath', () => {
   describe('ROOT', () => {
@@ -35,12 +36,9 @@ describe('CookiePath', () => {
       { input: '/foo/bar', expected: '/foo' },
       { input: '/foo/bar/', expected: '/foo/bar' },
       { input: '/foo', expected: '/' },
-    ])(
-      'parentPath of "$input" is "$expected"',
-      ({ input, expected }) => {
-        expect(CookiePath.parentPath(CookiePath.parse(input)!)).toBe(expected)
-      },
-    )
+    ])('parentPath of "$input" is "$expected"', ({ input, expected }) => {
+      expect(CookiePath.parentPath(CookiePath.parse(input)!)).toBe(expected)
+    })
 
     it('returns undefined for ROOT', () => {
       expect(CookiePath.parentPath(CookiePath.ROOT)).toBeUndefined()
@@ -66,6 +64,21 @@ describe('CookiePath', () => {
             CookiePath.parse(cookiePath)!,
           ),
         ).toBe(expectedValue)
+      },
+    )
+  })
+
+  describe('permute', () => {
+    it.each([...permutePathCases])(
+      'permute("$path") => $permutations',
+      ({ path, permutations }) => {
+        const parsed = CookiePath.parse(path)!
+        expect(CookiePath.permute(parsed)).toEqual([...permutations])
+        permutations.forEach((permutation) => {
+          expect(CookiePath.match(parsed, CookiePath.parse(permutation)!)).toBe(
+            true,
+          )
+        })
       },
     )
   })
--- a/lib/__tests__/cookiePath.spec.ts
+++ b/lib/__tests__/cookiePath.spec.ts
@@ -26,4 +26,22 @@ describe('CookiePath', () => {
       expect(CookiePath.parse(input)).toBeUndefined()
     })
   })
+
+  describe('parentPath', () => {
+    it.each([
+      { input: '/foo/bar/baz', expected: '/foo/bar' },
+      { input: '/foo/bar', expected: '/foo' },
+      { input: '/foo/bar/', expected: '/foo/bar' },
+      { input: '/foo', expected: '/' },
+    ])(
+      'parentPath of "$input" is "$expected"',
+      ({ input, expected }) => {
+        expect(CookiePath.parentPath(CookiePath.parse(input)!)).toBe(expected)
+      },
+    )
+
+    it('returns undefined for ROOT', () => {
+      expect(CookiePath.parentPath(CookiePath.ROOT)).toBeUndefined()
+    })
+  })
 })
--- a/lib/__tests__/cookiePath.spec.ts
+++ b/lib/__tests__/cookiePath.spec.ts
@@ -1,5 +1,6 @@
 import { describe, expect, it } from 'vitest'
 import { CookiePath } from '../cookie/cookiePath.js'
+import { defaultPathCases } from './data/defaultPathCases.js'
 
 describe('CookiePath', () => {
   describe('ROOT', () => {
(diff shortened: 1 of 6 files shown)


lib/cookie/cookiePath.ts
```diff
@@ -0,0 +1,112 @@
...
+ * and {@link CookiePath.defaultPath} for RFC 6265 §5.1.4 default path computation.
+ *
+ * @public
+ */
+export type CookiePath = string & { readonly [tag]: true }
+
+/**
+ * @public
+ */
+export const CookiePath = {
```
wjhsf: It seems likely to cause confusion that the exported _type_ `CookiePath` does not describe the shape of the exported _const_ `CookiePath`.

lib/cookie/cookiePath.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  /**
+   * Validates that the input starts with `/` and returns a {@link CookiePath},
+   * or `undefined` if the input is not a valid cookie path.
+   *
+   * @param input - the string to validate
+   */
+  parse(input: string): CookiePath | undefined {
+    if (!input.startsWith('/')) return undefined
+    return input as CookiePath
```
wjhsf:
```suggestion
    return input.startsWith('/') ? input as CookiePath : undefined
```
🤷

lib/cookie/cookiePath.ts
wjhsf: I know this is an old library, but exporting a wrapper object feels dated. I'd rather flatten it and export methods directly.

lib/cookie/cookiePath.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  /**
+   * Computes the default cookie path from a request URI path per
+   * {@link https://www.rfc-editor.org/rfc/rfc6265.html#section-5.1.4 | RFC 6265 §5.1.4}.
+   *
+   * @param path - the path portion of the request URI
+   */
+  defaultPath(path?: Nullable<string>): CookiePath {
+    const requestPath = CookiePath.parse(path ?? '') ?? CookiePath.ROOT
+    return CookiePath.parentPath(requestPath) ?? CookiePath.ROOT
```
wjhsf:
```suggestion
    if (!path) return ROOT
    const requestPath = parse(path)
    if (!requestPath) return ROOT
    return parentPath(requestPath) ?? ROOT
```
It feels like a lot of `??`. We don't need to call the functions when we're using fallback values, because we know the result. That being said, it's a micro-optimization and I think the suggestion is uglier. 😅

lib/cookie/cookiePath.ts
```diff
@@ -0,0 +1,112 @@
...
+
+      if (
+        reqPath.startsWith(cookiePath) &&
+        reqPath[cookiePath.length] === '/'
+      ) {
+        return true
+      }
+    }
+
+    return false
```
wjhsf:
```suggestion
    return `reqPath.startsWith(cookiePath) && (
      cookiePath[cookiePath.length - 1] === '/' // cookie path ends with /
      reqPath[cookiePath.length] === '/' // cookie path immediately followed by /
    )
```
`reqPath.indexOf(cookiePath) === 0` is the same thing as `reqPath.startsWith(cookiePath)`

```suggestion
    return `reqPath.startsWith(cookiePath) && (
      cookiePath.at(-1) === '/' // cookie path ends with /
      reqPath.at(cookiePath.length) === '/' // cookie path immediately followed by /
    )
```
Alternatively using `.at()` for fun modernness.

lib/cookie/cookiePath.ts
```diff
@@ -0,0 +1,112 @@
...
+   * The array is in longest-to-shortest order.
+   *
+   * @param path - a validated cookie path
+   */
+  permute(path: CookiePath): CookiePath[] {
+    if (path === '/') {
+      return [CookiePath.ROOT]
+    }
+    const permutations: CookiePath[] = [path]
+    let current: string = path
```
wjhsf:
```suggestion
    const permutations = [path]
    let current = path
```
The annotations aren't strictly necessary; omitting `string` for `current` makes it a `CookiePath`.

lib/cookie/cookiePath.ts
```diff
@@ -0,0 +1,112 @@
...
+    while (current.length > 1) {
+      const lindex = current.lastIndexOf('/')
+      if (lindex === 0) {
+        break
+      }
+      current = current.slice(0, lindex)
+      permutations.push(current as CookiePath)
```
wjhsf:
```suggestion
      current = current.slice(0, lindex) as CookiePath
      permutations.push(current)
```
If we make `current` a `CookiePath` then the assertion shifts location.

lib/__tests__/cookiePath.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+            CookiePath.parse(requestPath)!,
+            CookiePath.parse(cookiePath)!,
+          ),
+        ).toBe(expectedValue)
+      },
+    )
+  })
+
+  describe('permute', () => {
+    it.each([...permutePathCases])(
```
wjhsf:
```suggestion
    it.each(permutePathCases)(
```

lib/__tests__/cookiePath.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+      expect(CookiePath.parentPath(CookiePath.parse(input)!)).toBe(expected)
+    })
+
+    it('returns undefined for ROOT', () => {
+      expect(CookiePath.parentPath(CookiePath.ROOT)).toBeUndefined()
+    })
+  })
+
+  describe('defaultPath', () => {
+    it.each([...defaultPathCases])(
```
wjhsf:
```suggestion
    it.each(defaultPathCases)(
```

lib/__tests__/cookiePath.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+import { defaultPathCases } from './data/defaultPathCases.js'
+import { pathMatchCases } from './data/pathMatchCases.js'
+import { permutePathCases } from './data/permutePathCases.js'
+
+describe('CookiePath', () => {
+  describe('ROOT', () => {
+    it('equals "/"', () => {
+      expect(CookiePath.ROOT).toBe('/')
+    })
+  })
```
wjhsf, line 12: :100:

lib/__tests__/cookiePath.spec.ts
```diff
@@ -0,0 +1,85 @@
+import { describe, expect, it } from 'vitest'
+import { CookiePath } from '../cookie/cookiePath.js'
+import { defaultPathCases } from './data/defaultPathCases.js'
+import { pathMatchCases } from './data/pathMatchCases.js'
+import { permutePathCases } from './data/permutePathCases.js'
+
```
wjhsf, line 6: Weird to have some test cases in a file and some inlined.

lib/cookie/cookiePath.ts
colincasey: Good catch — this is resolved by flattening. There's no longer a `CookiePath` const at all; the module now exports the `CookiePath` type alongside standalone functions (`parse`, `match`, `defaultPath`, etc.), so the name only ever refers to the type. Consumers use `import * as CookiePath` to get the namespace.

colincasey: Agreed, done. Flattened to module-level function exports and dropped the wrapper object entirely.

lib/cookie/cookiePath.ts
colincasey: Kept the parens — prettier under our ESLint 10 config enforces `(input as CookiePath)`, so the unparenthesized form fails `npm run lint`.

lib/cookie/cookiePath.ts
colincasey: Kept the current form — since you flagged the suggestion as uglier and it's a micro-opt, the two-line `parse(...) ?? ROOT` / `parentPath(...) ?? ROOT` reads cleaner to me. Happy to revisit if you'd rather have the explicit early returns.

lib/cookie/cookiePath.ts
colincasey: Switched to `reqPath.startsWith(cookiePath)` — cleaner, thanks. Kept bracket indexing (`cookiePath[cookiePath.length - 1]`) over `.at()` to stay consistent with the rest of the file, but no strong opinion if you'd prefer `.at()`.

lib/cookie/cookiePath.ts
colincasey: Done — `current` is now `let current = path` and infers `CookiePath`.

lib/cookie/cookiePath.ts
colincasey: Done — assertion moved to the assignment: `current = current.slice(0, lindex) as CookiePath`.

lib/__tests__/cookiePath.spec.ts
colincasey: Done — dropped the spread, now `it.each(defaultPathCases)`.

lib/__tests__/cookiePath.spec.ts
colincasey: Done — dropped the spread here too, now `it.each(permutePathCases)`.

lib/__tests__/cookiePath.spec.ts
colincasey: The extracted files (`defaultPathCases`, `permutePathCases`, `pathMatchCases`) are the ones shared with the deprecated wrapper specs later in the stack — pulling them into data files lets both the `CookiePath.*` spec and the `defaultPath`/`pathMatch`/`permutePath` wrapper specs verify against the same corpus. The cases that stay inline (`parse`, `parentPath`) are unique to `CookiePath` and aren't reused, so there was no reason to extract them. I added header comments to each shared file noting this.

wjhsf (approved):

PR #599, opened 2026-04-27
refactor: deprecate path wrappers, delegate to CookiePath

colincasey:
## Summary

- Extracts test cases into shared fixtures under `lib/__tests__/data/`
- Deprecates `defaultPath`, `pathMatch`, and `permutePath` as thin wrappers delegating to the `CookiePath` const object
- `pathMatch` wrapper preserves backward compatibility for non-`/`-prefixed inputs via an equality fallback
- All existing behavior is preserved

Part 2 of 3 for #582. Stacked on `cookie-path-type` (#598).

Next: #cookie-path-internal-callers (migrates internal callers, exports `CookiePath`)

### DIFF
--- a/api/tough-cookie.api.md
+++ b/api/tough-cookie.api.md
@@ -149,7 +149,7 @@ export interface CreateCookieOptions {
     value?: string;
 }
 
-// @public
+// @public @deprecated
 export function defaultPath(path?: Nullable<string>): string;
 
 // @public
@@ -236,13 +236,13 @@ export interface ParseCookieOptions {
 // @public
 export function parseDate(cookieDate: Nullable<string>): Date | undefined;
 
-// @public
+// @public @deprecated
 export function pathMatch(reqPath: string, cookiePath: string): boolean;
 
 // @public
 export function permuteDomain(domain: string, allowSpecialUseDomain?: boolean): string[] | undefined;
 
-// @public
+// @public @deprecated
 export function permutePath(path: string): string[];
 
 // @public
--- a/lib/__tests__/defaultPath.spec.ts
+++ b/lib/__tests__/defaultPath.spec.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-deprecated */
 import { describe, expect, it } from 'vitest'
 import { defaultPath } from '../cookie/defaultPath.js'
 import { defaultPathCases } from './data/defaultPathCases.js'
--- a/lib/__tests__/defaultPath.spec.ts
+++ b/lib/__tests__/defaultPath.spec.ts
@@ -1,29 +1,12 @@
 import { describe, expect, it } from 'vitest'
 import { defaultPath } from '../cookie/defaultPath.js'
+import { defaultPathCases } from './data/defaultPathCases.js'
 
 describe('defaultPath', () => {
-  it.each([
-    {
-      input: null,
-      output: '/',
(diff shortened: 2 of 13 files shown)


lib/__tests__/defaultPath.spec.ts
```diff
@@ -1,29 +1,13 @@
...
 import { describe, expect, it } from 'vitest'
 import { defaultPath } from '../cookie/defaultPath.js'
+import { defaultPathCases } from './data/defaultPathCases.js'
 
 describe('defaultPath', () => {
-  it.each([
-    {
-      input: null,
-      output: '/',
+  it.each([...defaultPathCases])(
```
wjhsf:
```suggestion
  it.each(defaultPathCases)(
```

lib/cookie/permutePath.ts
```diff
@@ -9,21 +11,11 @@
...
-    if (lindex === 0) {
-      break
-    }
-    path = path.slice(0, lindex)
-    permutations.push(path)
-  }
-  permutations.push('/')
-  return permutations
+  const parsed = CookiePath.parse(path)
+  if (!parsed) return ['/']
```
wjhsf:
```suggestion
  if (!parsed) return [CookiePath.ROOT]
```
🤷

lib/__tests__/permutePath.spec.ts
```diff
@@ -1,29 +1,21 @@
...
 import { permutePath } from '../cookie/permutePath.js'
 import { pathMatch } from '../pathMatch.js'
+import { permutePathCases } from './data/permutePathCases.js'
 
 describe('permutePath', () => {
-  it.each([
-    {
-      path: '/',
-      permutations: ['/'],
+  it.each([...permutePathCases])(
```
wjhsf:
```suggestion
  it.each(permutePathCases)(
```

api/tough-cookie.api.md
```diff
@@ -122,6 +122,21 @@ export class CookieJar {
     toJSON(): SerializedCookieJar | undefined;
 }
 
+// @public
+export type CookiePath = string & {
+    readonly [tag]: true;
+};
+
+// @public (undocumented)
```
wjhsf: `(undocumented)` quoth the docs

wjhsf (approved):

lib/__tests__/defaultPath.spec.ts
colincasey: Done — dropped the spread, now `it.each(defaultPathCases)`.

lib/__tests__/permutePath.spec.ts
colincasey: Done — dropped the spread, now `it.each(permutePathCases)`.

lib/cookie/permutePath.ts
colincasey: This fallback changed — to stay bug-for-bug compatible with the pre-deprecation behavior, `permutePath` now falls back to the original algorithm for non-RFC inputs (e.g. `permutePath('noslash')`) rather than returning a single-element array. And since `CookiePath` is now internal, `ROOT` isn't exported anyway.

api/tough-cookie.api.md
colincasey: Resolved a different way — `CookiePath` is now marked `@internal`, so it's dropped from the public API report entirely rather than appearing as an undocumented public export.

PR #614, opened 2026-07-06
fix: avoid decoding URL authority part

wjhsf:
Decoding the URL can result in incorrect parsing in some cases.


colincasey (approved):
Looks good. Let's just add a decent PR description to this change.

PR #625, opened 2026-08-06
Replace tldts with tldkit

### LINKED ISSUE #624: Replace tldts with a smaller Public Suffix List package
Opening this before a pull request, per CONTRIBUTING.

tough-cookie installs 3,929 KB, and almost all of it is `tldts` plus `tldts-core` at 3,186.9 KB. Of that, about 1,809 KB is sourcemaps across 33 files, and the suffix trie ships several times over: as TypeScript source, as a `.d.ts`, once per module format under `dist/es6` and `dist/cjs`, and inlined again into three prebuilt bundles. The Public Suffix List is not what makes it big.

Nothing here is a criticism of `tldts`. It is maintained and fast, which is why #338 moved to it, and both of those still hold. This is only about what ends up on disk.

I wrote `tldkit` to package the same list differently: 140.9 KB installed, no runtime dependencies, ESM and CJS with types for both, and 645 KB of heap with the list resident against 1,140 KB. It exports `getDomain` with the same signature and defaults, so `lib/getPublicSuffix.ts` needs a one line import change. Warm throughput is the same. The first lookup costs 0.77 ms against 0.33 ms, because it expands its packed dataset on first use rather than at module load.

It scores 76/78 on the publicsuffix.org conformance suite, the same score and the same two failing cases as `tldts`, and was checked against `tldts` over 105,146 cases covering every rule in the list under both `allowPrivateDomains` settings. The dataset is generated from publicsuffix.org and committed, so install touches no network, and a weekly job regenerates it and opens a pull request when the list moves.

I have the change working locally. Your full suite passes, 784 tests, along with lint, format, `attw --pack` and `api-extractor`.

One thing you would be taking on. The two libraries disagree on 21 hostnames. Seven PSL wildcard rules have a deeper wildcard sibling, for example `*.run.app` alongside `*.mtls.run.app`. For `mtls.run.app` itself, `tldkit` returns no registrable domain and `tldts` returns `run.app`, so `Domain=mtls.run.app` would be rejected as a public suffix under `tldkit` and accepted under `tldts`. Subdomains of those hosts agree. `tldkit` follows the publicsuffix.org prevailing rule algorithm here. The rest sit under `customer-oci.com`, `snowflake.app`, `firenet.ch`, `futurecms.at` and `cloud.int.apple`. I have not reported this to `tldts`.

Pull request to follow.

https://github.com/Damin3927/tldkit

Damin3927:
Closes #624.

Swaps the one runtime dependency, `tldts`, for `tldkit`. The only source change is the import in `lib/getPublicSuffix.ts`. `tldkit` exports `getDomain` with the same signature and defaults, so the call site is unchanged.

The reason is install size. `tldts` plus `tldts-core` is 3,186.9 KB installed, nearly all of tough-cookie's 3,929 KB tree. `tldkit` is 140.9 KB. About 1,809 KB of the `tldts` tree is sourcemaps, and the suffix trie ships several times over. Heap with the list resident goes from 1,140 KB to 645 KB, which matters for test runners that fork per file. Warm throughput is the same, but `tldkit` expands its packed dataset on first lookup rather than at module load, so the first call costs 0.77 ms against 0.33 ms.

`tldkit` is MIT, has no runtime dependencies, ships ESM and CJS with types for both, Node 16+, and commits its dataset so install hits no network. It scores 76/78 on the publicsuffix.org conformance suite, the same two failures as `tldts`, and matches it across 105,146 generated cases under both `allowPrivateDomains` settings.

One caveat, also in the issue. The two disagree on 21 hostnames, from 7 PSL wildcard rules that have a deeper wildcard sibling, such as `*.run.app` and `*.mtls.run.app`. For `mtls.run.app` itself, `tldkit` returns no registrable domain and `tldts` returns `run.app`, so `Domain=mtls.run.app` is rejected as a public suffix under `tldkit` and accepted under `tldts`. Subdomains of those hosts agree. `tldkit` follows the publicsuffix.org algorithm here. The others sit under `customer-oci.com`, `snowflake.app`, `firenet.ch`, `futurecms.at` and `cloud.int.apple`. Not reported upstream.

`npm test` passes locally (784 tests), as do lint, format, `attw --pack` and `api-extractor`.

https://github.com/Damin3927/tldkit

### DIFF
--- a/lib/getPublicSuffix.ts
+++ b/lib/getPublicSuffix.ts
@@ -1,4 +1,4 @@
-import { getDomain } from 'tldts'
+import { getDomain } from 'tldkit'
 
 // RFC 6761
 const SPECIAL_USE_DOMAINS = ['local', 'example', 'invalid', 'localhost', 'test']
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "6.0.2",
       "license": "BSD-3-Clause",
       "dependencies": {
-        "tldts": "^7.0.5"
+        "tldkit": "^1.0.2"
       },
       "devDependencies": {
         "@arethetypeswrong/cli": "^0.18.2",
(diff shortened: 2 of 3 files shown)


wjhsf:
Thanks for the contribution! However, while `node_modules` bloat is a notorious problem across npm, package size in this case is not a major concern for us. Most of the files, as you mentioned, are unused at runtime; the total size of the actual JS that gets executed is about 225 KB. In addition, `tldts` is a widely used, actively maintained project with a history of reliability. Given the number of dependents that this project has, reputation is an important factor in how we select our dependencies.

If you'd like to improve on the size `tldts` package, I'd recommend contributing directly to that project, rather than forking it.

PR #637, opened 2026-09-03
Fix Max-Age serializing as the literal word Infinity/-Infinity

afonsojanu:
`setMaxAge()` accepts `Infinity` and `-Infinity` and stores them as the strings `'Infinity'` / `'-Infinity'` on the cookie (there's a comment explaining this is so the value survives a JSON round trip, since actual `Infinity` doesn't serialize). `toString()` was supposed to skip the Max-Age attribute for both of those, same as it does for `expires: 'Infinity'`, but it only actually does that for the positive case:

```js
const c = new Cookie({ key: 'a', value: 'b' })
c.setMaxAge(-Infinity)
c.toString() // 'a=b; Max-Age=-Infinity'
```

The reason is a loose-equality accident: the check was `this.maxAge != Infinity`, and in JS `'Infinity' == Infinity` is true (string gets coerced to a number) while `'-Infinity' == Infinity` is false, so the negative case slipped through to the serialization branch untouched.

`Max-Age=-Infinity` isn't something any real HTTP client can parse — RFC 6265's delta-seconds is a plain signed integer, no word tokens. A cookie in this state represents "expire right now," so the fix emits `Max-Age=0` instead, which is the normal, widely understood way to tell a client to drop a cookie immediately.

Also switched the Infinity check itself to a strict string comparison rather than relying on the `==` coercion, since that's what was masking the bug.

Added two tests to `cookieToString.spec.ts` covering both the Infinity (still omitted) and -Infinity (now `Max-Age=0`) cases. Ran the full suite (786 tests) and lint, both clean.

### DIFF
--- a/lib/__tests__/cookieToString.spec.ts
+++ b/lib/__tests__/cookieToString.spec.ts
@@ -51,4 +51,17 @@ describe('Cookie.toString()', () => {
     })
     expect(cookie.toString()).toBe('a=b; Expires=Fri, 10 Sep 2010 10:10:10 GMT')
   })
+
+  it('should omit Max-Age entirely when it has been set to Infinity', () => {
+    const cookie = new Cookie({ key: 'a', value: 'b' })
+    cookie.setMaxAge(Infinity)
+    expect(cookie.toString()).toBe('a=b')
+  })
+
+  it('should output a real Max-Age, not the word "-Infinity", for a cookie set to expire immediately', () => {
+    const cookie = new Cookie({ key: 'a', value: 'b' })
+    cookie.setMaxAge(-Infinity)
+    expect(cookie.maxAge).toBe('-Infinity')
+    expect(cookie.toString()).toBe('a=b; Max-Age=0')
+  })
 })
--- a/lib/cookie/cookie.ts
+++ b/lib/cookie/cookie.ts
@@ -750,8 +750,11 @@ export class Cookie {
       }
     }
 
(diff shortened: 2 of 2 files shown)


afonsojanu:
I just signed up!

wjhsf:
`Infinity` is a special value, but `-Infinity` is not. As mentioned, it is an invalid value, so it should be treated the same as other invalid values. Other invalid values have no special handling; `c.setMaxAge(-1)` is accepted and not modified, as is `c.setMaxAge('foo')`. Thus, `c.setMaxAge(-Infinity)` should result in `c.maxAge === -Infinity` and `c.toString()` should include `Max-Age: -Infinity`. The difference from the current behavior is solely that `c.setMaxAge(-Infinity)` currently gets coerced to a string, `c.maxAge === '-Infinity'`, but it should not.

afonsojanu:
You're right, I looked at this again and there wasn't actually a bug here. `-Infinity` is already a documented sentinel in this file (see the `maxAge?: number | 'Infinity' | '-Infinity' | null` type, and how `isPersistent`-ish validation and `expiryTime()` both treat the string '-Infinity' as its own case rather than an error). The loose-equality check I flagged happens to produce the right output for both special values by coincidence of how JS compares a numeric Infinity against those strings, so nothing was actually falling through by accident.

Emitting `Max-Age=0` instead would've been a behavior change dressed up as a bug fix, not something this PR should be doing. Closing it, thanks for catching this before it went further.

